PR #3, opened 2025-05-21
SYS-1835: support for viewing, editing, and adding individual items

ztucker4:
Implements [SYS-1835](https://uclalibrary.atlassian.net/browse/SYS-1835)

Allows users to add new items, and to view and edit existing records.

All three templates (view, edit, and add) use a button with some simple JavaScript (now in `main.js` and imported through a `base` template) to show/hide "advanced" metadata fields. For now, "basic" fields include columns up to Title in the original spreadsheet. 

Add item interface: http://127.0.0.1:8000/add_item/ 
View item example: http://127.0.0.1:8000/view_item/2348/ (record should exist if using data loaded from DL Sheet)
Edit item example: http://127.0.0.1:8000/edit_item/2348/ 

Note/Question: `add_item.html` and `edit_item.html` are very similar. The only differences are the contents of the `<h1>` titles and the text on `submit` buttons. It seems like there should be a way to reuse code here, but I haven't figured it out.

[SYS-1835]: https://uclalibrary.atlassian.net/browse/SYS-1835?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


akohler:
@ztucker4 This looks great - thank you for your quick work on this!  I'm willing to merge as-is, but you asked:

> Note/Question: add_item.html and edit_item.html are very similar. The only differences are the contents of the `<h1>` titles and the text on submit buttons. It seems like there should be a way to reuse code here, but I haven't figured it out.

I think an easy way to address this is to use the `context` when rendering the template from the view(s).  For example, you could set two additional keys like this (untested, with a hypothetical `add_edit_item.html` combined template):

```
return render(request, "add_edit_item.html", {"form": form, "title": "Add Item", "button_text": "Add Item"})
```

Then in the template, just use the relevant variable value instead of the current constant text.

Can you give that a try?
Thanks --Andy

ztucker4:
Thanks Andy! I made the suggested changes, and now adding and editing is handled with the single `add_edit_item.html` template with context passed from the views.

ftva_lab_data/views.py
```diff
@@ -1,3 +1,75 @@
...
+            messages.success(request, "Item added successfully!")
+            return redirect("view_item", item_id=form.instance.id)
+
+        else:
+            # Handle form errors, if needed
+            messages.error(request, "Please correct the errors below.")
+            return render(
+                request,
+                "add_edit_item.html",
+                {"form": form, "title": "Add Item", "button_text": "Add Item"},
```
akohler, line 22: Since the same context (the same collection of key/value pairs) is passed in multiple cases (here and line 30), please define a `context` dictionary and reference it in both places.

ftva_lab_data/views.py
```diff
@@ -1,3 +1,75 @@
...
+        if form.is_valid():
+            form.save()
+            messages.success(request, "Item updated successfully!")
+            return render(request, "view_item.html", {"item": item})
+        else:
+            messages.error(request, "Please correct the errors below.")
+            return render(
+                request,
+                "add_edit_item.html",
+                {
```
akohler, line 49: Same as in `add_item`, please define a `context` dict for these key/values and refer to it here and in line 62.

akohler (changes requested):
@ztucker4 The template unification looks and works well, thanks.
Just 2 more changes, to use `context` variables in the views where the same collection of keys/values are passed, then I'll merge.

ftva_lab_data/views.py
ztucker4: Context dictionaries added to the `add_item` and `edit_item` views. Ready for re-review!

akohler (approved):
@ztucker4 Thanks for the context changes.  This all works, merging now.

PR #4, opened 2025-05-22
SYS-1836: add ability to view and search items

henry-r18:
Implements [SYS-1836](https://uclalibrary.atlassian.net/browse/SYS-1836)

### Acceptance criteria

- ✅ A form allows basic searching of DL sheet data in Django
  - I went with a table, but same effect
- ✅ By default, search for the given term as a substring in all fields (same as basic Google / browser “Find in sheet” functionality)
- ✅ Allow users to target specific fields / groups of fields, like “Title” and “File/Folder” fields.
- ❌  More advanced features, like filtering by format & similar metadata, may be added later but is out of scope for now.
  - Haven't yet implemented these advanced search features

### New library
I used [HTMX](https://htmx.org/), a small but mighty JavaScript library, to do things like reactively re-render the table part of the main template without reloading the whole page. It also addresses the issue raised during standup about keeping the client state in sync with Django. It may prove useful later on for similar things, so though it's not pure Django, I figured it was worth it to include it. I opted to include it as a local static file, rather than via a CDN, so we don't have to worry as much about it changing.

### Routes, views, and templates
This PR adds two new routes: one for the initial page at the base URL and another at `render_table/` that's used to render the table markup in `partials/table.html`. I opted to keep the table as a partial to make it easier for HTMX to inject it into the page on load and reactively when the search controls change.

There are two view functions: a simple one to render `view_items.html` and a more complicated one that handles two types of search—both column-scoped and cross-column—as well as pagination.

### Testing
I didn't create any unit tests at this time. I manually tested the search features and pagination on the table, as well as the links to the individual item views in each row. Data persists when saved via the editing forms, but I have not yet added a way to navigate back to the table from the form. If you use the back button in the browser, you may need to reload the home page for the table to refresh with the data.

screen_cap


ftva_lab_data/templates/view_items.html
akohler, line 1: Please rename this file, to help distinguish it from `view_item.html`.  The purpose of this template is search results, so  `search_results.html` - that's what we've used in other projects.

ftva_lab_data/templates/view_items.html
```diff
@@ -0,0 +1,30 @@
...
+    <div class="d-flex w-50 gap-2">
+        <select name="column" class="form-select">
+            <option value="">All columns</option>
+            {% for field, label in columns %}
+            <option value="{{ field }}">{{ label }}</option>
+            {% endfor %}
+        </select>
+
+        <input id="search-input" class="form-control" type="text" name="search" placeholder="Enter search term...">
+    </div>
```
akohler, line 21: Can you add a "reset" button, or an `X` control of some sort?  Not critical and it may not fit with the HTML aesthetic, but search interfaces generally have a UI element to at least clear the search box.

ftva_lab_data/templates/view_items.html
```diff
@@ -0,0 +1,30 @@
...
+</form>
+
+
+
+<!-- On page load, HTMX triggers GET request to 'render_table'
+and inserts it within div -->
+<div id="table-container" class="p-2" hx-get="{% url 'render_table' %}" hx-trigger="load"></div>
+
+{% endblock %}
```
akohler, line 30: Please add a linefeed at the end of the last line, so the file ends with an empty line 31.

ftva_lab_data/views.py
```diff
@@ -57,3 +62,54 @@ def view_item(request, item_id):
     item = SheetImport.objects.get(id=item_id)
 
     return render(request, "view_item.html", {"item": item})
+
+
+def view_items(request: HttpRequest) -> HttpResponse:
```
akohler, line 67: This view should be renamed to better show its purpose, corresponding to the template rename.

ftva_lab_data/views.py
```diff
@@ -57,3 +62,54 @@ def view_item(request, item_id):
...
+    return render(request, "view_items.html", context={"columns": COLUMNS})
+
+
+def render_table(request: HttpRequest) -> HttpResponse:
```
akohler, line 71: I'd probably rename this too, to be more specific.  `render_search_results` or `render_search_table`, probably.

ftva_lab_data/templates/partials/table.html
akohler, line 1: Combined with the other renames, probably rename this to `partials/search_results.html` (or `partials/search_results_table.html`), consistent with changes to the view names.

ftva_lab_data/views.py
```diff
@@ -57,3 +62,54 @@ def view_item(request, item_id):
...
+            # Scoped search to selected column
+            items = items.filter(**{f"{column}__icontains": search})
+        else:
+            # General CTRL-F-style search across all configured fields
+            query = Q()  # start with empty Q() object, always True
+            for field in display_fields:  # then add queries for all valid fields
+                query |= Q(**{f"{field}__icontains": search})
+            items = items.filter(query)
+
+    paginator = Paginator(items, 10)  # TODO: make configurable
```
akohler, line 96: The default of 10 seems reasonable.  If it's easy to add a UI element to let the user change this, go ahead... though if they'd have to change it again with every session, I'd rather not get into providing personal settings persistence at this point.
Otherwise, this `TODO` can be removed :)

akohler (changes requested):
@henry-r18 This is excellent!  It looks great, performs well, and really exceeds my expectations for this functionality.  HTMX is an excellent choice.  I expect we'll have you start retrofitting other applications with this :).  I also learned about Django's `Paginator`, which I'd not seen before.

I am requesting several small changes, mostly renaming templates & views to better reflect their specific purposes.  I also think it's worth adding (source) inventory number to the search interface at this point.   Once these tweaks are made, I'll merge.

Thanks --Andy

PR #5, opened 2025-05-27
SYS-1840: reorganize item edit UI

henry-r18:
Implements [SYS-1840](https://uclalibrary.atlassian.net/browse/SYS-1840)

#### Acceptance criteria
- ✅ Redesign the edit_item screen to make better use of wide-screen monitors
- ✅ Try to fit the first 9 fields (hard_drive_name through source_inventory_number) into the top half of the screen, via 2-3 columns
  - Using Bootstrap's grid system, I used 3 columns for the Basic Fields section
- ✅ The remaining fields could be handled in 4-5 columns, as most have short values
  - I used 4 columns for the remaining Advanced Fields section

#### Other changes
I moved the buttons from below to above the form, to avoid the `Save` button jumping to the bottom when the `Advanced Fields` section is revealed. I also changed the buttons to vanilla HTML button elements with Bootstrap classes, rather than `django-bootstrap5` template imports, as I couldn't figure out how to add HTMX attributes to `{% bootstrap_button %}`. It didn't appear to be able to expand the HTMX attribute, which uses `hx-` as a prefix, as a keyword argument for the underlying method call.

I also changed the enclosing tag around the two field sections (Basic Fields and Advanced Fields) from `<div>` to `<fieldset>`, just to be a bit more semantic.

The changes to the JavaScript file just wraps the toggle for Show/Hide Advanced Fields in a function and refactors it a bit.

### DIFF
--- a/static/js/main.js
+++ b/static/js/main.js
@@ -1,15 +1,7 @@
-let toggleAdvancedFields = document.getElementById("toggle-advanced-fields");
-if (toggleAdvancedFields) {
-  toggleAdvancedFields.addEventListener("click", function () {
-    const advancedFields = document.getElementById("advanced-fields");
-    if (advancedFields.style.display === "none") {
-      advancedFields.style.display = "block";
-      this.textContent = "Hide Advanced Fields";
-    } else {
-      advancedFields.style.display = "none";
-      this.textContent = "Show Advanced Fields";
-    }
-  });
+function toggleAdvancedFields(el) {
+  const advancedFields = document.getElementById("advanced-fields");
+  const isHidden = advancedFields.hidden = !advancedFields.hidden;
+  el.textContent = isHidden ? "Show Advanced Fields" : "Hide Advanced Fields";
 }
 
 function clearSearchInput() {
--- a/ftva_lab_data/templates/add_edit_item.html
+++ b/ftva_lab_data/templates/add_edit_item.html
@@ -5,73 +5,93 @@
 {% block content %}
 
 {% if messages %}
-    <div class="alert alert-success">
-        {% for message in messages %}
-            {{ message }}
-        {% endfor %}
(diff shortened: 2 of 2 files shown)


static/js/main.js
```diff
@@ -1,15 +1,7 @@
...
-      this.textContent = "Hide Advanced Fields";
-    } else {
-      advancedFields.style.display = "none";
-      this.textContent = "Show Advanced Fields";
-    }
-  });
+function toggleAdvancedFields(el) {
+  const advancedFields = document.getElementById("advanced-fields");
+  const isHidden = advancedFields.hidden = !advancedFields.hidden;
+  el.textContent = isHidden ? "Show Advanced Fields" : "Hide Advanced Fields";
```
akohler, line 4: @henry-r18 I like the changes to this, making it cleaner and more modern.  It works well on the `edit_item` screen... but unfortunately, this change has broken the toggle on the `view_item` screen.

For example, see http://127.0.0.1:8000/view_item/6/
Clicking "Show Advanced Fields" now does nothing.

I think the function is fine, but the `view_item` template should be updated as needed to work with the new javascript.

akohler (changes requested):
@henry-r18 The redesigned "Edit Item" screen is clean, looks good and functions correctly - this is a good improvement.
As noted separately, the javascript refactoring works in the "edit" context, but is now broken in the "View Item" context.  Please fix that, then I'll merge.
Thanks --Andy

akohler (approved):
@henry-r18 This all looks great now.  I like the revised Edit screen, and the corresponding changes to the View screen.  Toggling advanced field display on both screens works correctly.  I also like the extra buttons for "Back to Search" and "Cancel".  Merging now, thanks.

@ztucker4 FYI.

PR #6, opened 2025-05-29
SYS-1842: Allow assignment of records to users

ztucker4:
Implements [SYS-1842](https://uclalibrary.atlassian.net/browse/SYS-1842)

Allows users to assign and unassign records to users from the search results page.

To test, run the new migration: `python manage.py migrate`, then view the search results page in the UI: http://127.0.0.1:8000/

You should be able to:
* Assign records to users by checking the desired checkboxes, selecting a user from the dropdown, and clicking "Assign selected".
* Unassign records by using the "Unassign" value in the "Assign to..." dropdown.
* Select all visible records using the top checkbox, and assign them as desired.
* After records are assigned or unassigned, see a success message at the top of the page.
* Maintain table state after assigning records. Try different combinations of search column, search text, and page. These values should remain the same after submitting the user assignment.

Notes on code changes:
* Messages have been pulled out from templates into a new `messages.html` partial. This was needed to deal with displaying messages after `HX-Request` requests on the search results page, and also reduces code duplication.
* HTMX attributes I learned about when implementing this ticket: [hx-target](https://htmx.org/attributes/hx-target/), [hx-swap-oob](https://htmx.org/attributes/hx-swap-oob/)


ftva_lab_data/views_utils.py
```diff
@@ -0,0 +1,14 @@
...
+from django.db import models
+
+
+def get_field_value(obj: models.Model, field_name: str) -> Any:
+    """
+    Helper function to get the value of a field from an object.
+    Useful for double underscore prefixed fields.
+    """
+    for part in field_name.split("__"):
+        obj = getattr(obj, part, None)
```
akohler, line 11: I found it hard to follow this implementation, though it does work as expected.  Part of it is the reuse of `obj`: I understand that for related fields like `assigned_user__username`, the loop first retrieves the `User` object, assigns that value to `obj`, then ultimately gets `User.username`.  `None` is used as a guard value, which ultimately gets discarded in favor of an empty string if there's no data, or if the method is called with a `field_name` that doesn't exist on a model.

It's also not clear to me that this would work correctly with deeper nesting of foreign keys than just the 2 levels we're currently using - it might, but the current implementation is harder to envision, for me at least.

I think a cleaner implementation uses recursion.  I found this, which I did not write but I've tried it out and it seems to match the functionality of your implementation.  The recursion makes it clearer that it should work to any depth of FK relations, and it avoids the `None` vs empty string complexity.
https://stackoverflow.com/a/74613925

Please take a look, and if you agree, please update your implementation to use this approach (with attribution, if you use it largely as-is, or rewrite to use a similar approach).

I hope this makes sense; happy to discuss further.  Thanks.

akohler (changes requested):
@ztucker4 This works correctly, but I think needs two changes (other than a minor issue with `get_user_model()`):
1. I think the implementation of `get_field_value()` should be changed.  See detailed comments on that function.  This is also a good candidate for some tests.
2. The unchecked checkboxes are very hard to see.  I tried in both Firefox (my norm, with and without dark theme enabled) and in Edge (rarely used, no extensions or customizations), with the same visual results.  Unchecked boxes are nearly invisible, though they show very clearly once checked.  Here's a screenshot of what I see:

ftva_django_faint_assign_boxes

Thanks --Andy

ztucker4:
Hi @akohler, this is ready for re-review:

* Updated `SheetImport` model to use `settings.AUTH_USER_MODEL`. This does not require a new/updated migration - the existing one in this branch shows that Django is already using this value.
* Rewrote `get_field_value()` using the recursive approach you linked, with more helpful comments in the docstring.
* Added tests for `get_field_value()`. Run with `python manage.py test`.
* Added custom CSS to make checkbox borders more visible. I think this should be ok contrast in both light and dark mode.
image 
image

akohler (approved):
@ztucker4 Thanks for making the code changes.  The code looks good to me, functions as expected, all tests pass.
I'm still seeing minimal contrast with the checkboxes on my system, in both Firefox (with and without dark theme enabled there) and in Edge (no special theme or extensions).  I suspect this is related to my Windows 10 system-wide dark mode setting.  Since your screenshots show good contrast, I'll merge this, but it may need more work later depending on feedback from staff.
Merging now, thanks --Andy

PR #7, opened 2025-05-29
SYS-1841: restrict editing, add login, and add user-creation script

henry-r18:
Implements [SYS-1841](https://uclalibrary.atlassian.net/browse/SYS-1841)

| **Acceptance Criteria** | **Notes** |
|--------|--------|
| ✅ There is an editors groups | The `editors` group is defined in the `groups_and_permissions.json` fixture, with appropriate permissions for adding and editing `SheetImport` objects assigned to it |
| ✅ The group and permission definitions are exported to a fixture | The `groups_and_permissions.json` fixture can be loaded via `python manage.py loaddata groups_and_permissions.json` |
| ✅ All editing functionality is restricted to users in the editors group | This is accomplished using the `@permission_required` decorator on the `add_item` and `edit_item` views |
| ✅ All other functionality within the application is restricted to logged-in users | This is accomplished using the `@login_required` decorator on all views. A template for the login page is also included. |
| ✅ Write a script to create users from [List of Django Users](https://docs.google.com/spreadsheets/d/1EuAZnnNwsqXpa7FH20v4ruzkSntFioBGlyKwMLfbMYw/edit?gid=1630856852#gid=1630856852) | A new management command is defined to handle this. Download a copy of the spreadsheet as an XLSX file, then run `python manage.py create_users_from_sheet -f {SPREADSHEET}.xlsx`.<br /><br />As is, this script creates users, then sets their passwords with `user.set_unusable_password()`. I included an option `--email_users` that I figured could be used later to handle sending users a one-time link to reset their password, when the users are first created. |

### Tests
I wrote 5 tests to confirm that only users with the correct permissions, by way of their inclusion in the `editors` group, can add and edit items.

Run tests with `python manage.py test` after loading the `groups_and_permissions.json` fixture.

### Note about `view_item.html` template
To conditionally render the "Edit This Record" button only for users with suitable permissions, I settled on using `{% if perms.ftva_lab_data.change_sheetimport %}`, as documented [here](https://docs.djangoproject.com/en/5.1/topics/auth/default/#permissions). I figured this was more straightforward than handling the logic in the view or in a helper function.

### Note about `add_edit_item.html` template
Since the add and edit forms are the same, we're using a single template for the two views, `add_item` and `edit_item`. There was an issue accessing the `add_item` URL, because of the Cancel button I had added. I addressed this issue by making the `href` for the link conditionally point to the `view_item` URL if the template is being rendered from the `edit_item` view with an `item` defined in the context, or point to the `search_result` URL otherwise.

[SYS-1841]: https://uclalibrary.atlassian.net/browse/SYS-1841?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/README.md
+++ b/README.md
@@ -111,6 +111,18 @@ Certain views within the application are restricted to users with appropriate pe
 
 ```python manage.py loaddata groups_and_permissions.json```
 
+#### Loading users
+
+FTVA staff provided a list of users in a Google Sheet. Ask a teammate for the sheet, if need be. Users can be loaded from the sheet using the following management command:
+
+```python manage.py create_users_from_sheet -f {SPREADSHEET_PATH}.xlsx```
+
+The script accepts two command line arguments:
+
+- `-f` (or `--file_name`) _required_: path to an Excel (XLSX) export of the FTVA Google Sheet containing user data
+- `--email_users`: whether to email users with a link to set their passwords
+  - NOTE: the emailing logic is not currently implemented. This feature will be added later.
+
 ### Logging
 
 Basic logging is available, with logs captured in `logs/application.log`.  At present, logs from both the custom application code and Django itself are captured.
--- a/README.md
+++ b/README.md
@@ -105,6 +105,12 @@ The container runs via `docker_scripts/entrypoint.sh`, which
 
    ```Installed 26653 object(s) from 1 fixture(s)```
 
+#### Loading group and permission definitions
+
+Certain views within the application are restricted to users with appropriate permissions. Corresponding group and permission definitions are included in the `groups_and_permissions.json` fixture.  This can be loaded using:
+
+```python manage.py loaddata groups_and_permissions.json```
+
 ### Logging
 
 Basic logging is available, with logs captured in `logs/application.log`.  At present, logs from both the custom application code and Django itself are captured.
--- a/ftva_lab_data/templates/add_edit_item.html
+++ b/ftva_lab_data/templates/add_edit_item.html
@@ -18,7 +18,9 @@
             <button class="btn btn-primary" type="submit" form="item_form">{{ button_text }}</button>
         </div>
 
-        <a class="btn btn-outline-danger" href="{% url 'view_item' item.id %}">Cancel</a>
+        <a class="btn btn-outline-danger"
+            href="{% if item %}{% url 'view_item' item.id %}{% else %}{% url 'search_results' %}{% endif %}">
+            Cancel</a>
     </div>
 
     <form name="item_form" id="item_form" method="post" enctype="multipart/form-data">
--- a/ftva_lab_data/tests.py
+++ b/ftva_lab_data/tests.py
@@ -1,6 +1,8 @@
-from django.test import TestCase
+from django.test import TestCase, Client
 from ftva_lab_data.models import SheetImport
 from django.contrib.auth import get_user_model
+from django.contrib.auth.models import User, Group, Permission
+from django.urls import reverse
 from ftva_lab_data.views_utils import get_field_value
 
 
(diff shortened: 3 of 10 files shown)


ftva_lab_data/templates/add_edit_item.html
```diff
@@ -22,7 +22,9 @@ <h1>{{ title }}</h1>
             <button class="btn btn-primary" type="submit" form="item_form">{{ button_text }}</button>
         </div>
 
-        <a class="btn btn-outline-danger" href="{% url 'view_item' item.id %}">Cancel</a>
+        <a class="btn btn-outline-danger"
+            href="{% if item %}{% url 'view_item' item.id %}{% else %}{% url 'search_results' %}{% endif %}">
```
henry-r18, line 26: As noted in my main comment, this is to get the `add_item` view to render correctly.

henry-r18:
@akohler I re-exported the group and permission fixture using `python manage.py dumpdata --natural-foreign --indent 2 auth.group auth.permission > ftva_lab_data/fixtures/groups_and_permissions.json`. I am hoping that using the `--natural-foreign` flag will sidestep the foreign key issue. After dropping my DB, restarting, and running `loaddata`, I get:

`Installed 29 object(s) from 1 fixture(s)`

The fixture now includes only the `editors` group; I see what you mean about the `view_only` group. I went ahead and adjusted the `create_users_from_sheet.py` script accordingly, so that only users on the "editors" sheet will get added to the group, and users on other sheets will have no group assignment.

Thank you! --Henry

ftva_lab_data/tests.py
```diff
@@ -1,3 +1,112 @@
...
+
+        # Create two users, one to add to editors group and one with no group
+        cls.unauthorized_user = User.objects.create_user(
+            username="unauthorized", password="testpassword"
+        )
+        cls.authorized_user = User.objects.create_user(
+            username="authorized", password="testpassword"
+        )
+
+        # Create editors group and add permissions to add and edit SheetImport
```
akohler, line 27: Since the editors group is now defined in a fixture, load the fixture and use that definition instead of creating the group here in the tests.  That is different from how we did it in the OH Staff UI project tests... but those tests just checked group membership, rather than actual permissions as you do.  Your overall approach to permissions is better than we did in OH, so the tests can be better too :)

README.md
```diff
@@ -105,6 +105,12 @@ The container runs via `docker_scripts/entrypoint.sh`, which
 
    ```Installed 26653 object(s) from 1 fixture(s)```
 
+#### Loading group and permission definitions
+
+Certain views within the application are restricted to users with appropriate permissions. Corresponding group and permission definitions are included in the `groups_and_permissions.json` fixture.  This can be loaded using:
+
+```python manage.py loaddata groups_and_permissions.json```
+
 ### Logging
```
akohler, line 114: Please add documentation to the README for loading users.

akohler (changes requested):
@henry-r18 Thanks for updating the fixture.  I confirmed I can load (and reload, without duplication) this data now.  All tests pass.  I logged in as users both in and out of the Editors group and confirmed the UI behaves as expected, and I got 403 when trying to access unauthorized URLs directly.

I've requested a couple more small changes.  Also, since we want the permission definitions loaded consistently, please add the fixture to `docker_scripts/entrypoint.sh`, after `sample_data.json` is loaded.

Thanks --Andy

henry-r18:
@akohler I rebased my local branch on top of Zoe's changes, cured merge conflicts, then force-pushed here. I think I may have done git correctly, for once! Please let me know if I fumbled it though.

My most recent commit includes the changes to `entrypoint.sh`, `create_users_from_sheet.py`, `tests.py`, and `README` that you requested. All unit tests (9 of them now) are passing for me, and everything appears to be working as expected.

Best,
Henry

akohler (approved):
@henry-r18 Changes all look good to me.  Data loads (or reloads) via fixtures as expected, tests all pass, code looks correct, and everything works as expected within the UI.
I did have to delete my local branch, which had conflicts when I pulled after your force push, but that's OK - git just being git, nothing wrong on your end.
Merging now.  @ztucker4 FYI.
Thanks --Andy

henry-r18:
Thanks @akohler!

PR #8, opened 2025-05-30
SYS-1864: pagination bug fix

ztucker4:
Fixes [SYS-1864](https://uclalibrary.atlassian.net/browse/SYS-1846)

The bulk of this fix is in lines 43 and 49 of `search_results_table.html`, and line 141 of `views.py`. These changes allow the pagination buttons to pass the name of the column to be searched, as well as the search term and page number.

To test in browser, attempt to replicate the scenario [described in the ticket](https://uclalibrary.atlassian.net/browse/SYS-1846) (search in a column, page back and forth) and observe that the results do not change.

There is a new test case here, which does usefully test search results split across two pages. However, it does not simulate the results of actually clicking on the buttons in the UI - instead, requests with specific parameter values are sent manually. So this test doesn't actually correspond to the bug that was fixed here. If we want to test the actual button-click experience, I'm pretty sure this could be done with `BeautifulSoup` to get URL from the button. Let me know if you think this would be worth it.

I also renamed the variable `column` to `search_column` in templates, views, and JS to reduce (my own) confusion with the unrelated `COLUMNS`.

And finally, I included the new custom CSS file that should have been done with SYS-1842. Checkboxes should now be properly outlined.

### DIFF
--- a/ftva_lab_data/templates/partials/search_results_table.html
+++ b/ftva_lab_data/templates/partials/search_results_table.html
@@ -40,13 +40,13 @@
 <div class="d-flex justify-content-center align-items-center gap-2">
 
     <button class="btn btn-primary" {% if page_obj.has_previous %}
-        hx-get="{% url 'render_table' %}?page={{ page_obj.previous_page_number }}&search={{ search }}"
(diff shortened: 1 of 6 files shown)


ztucker4:
And sorry for the post-review-request commit - I cleaned up some misleading comments in the test file.

PR #12, opened 2025-06-04
SYS-1849: Add item statuses

ztucker4:
Implements [SYS-1849](https://uclalibrary.atlassian.net/browse/SYS-1849)

Adds a new model, `ItemStatus`, which is used as a `ManyToManyField` by the `SheetImport` model.

To test, restart containers to apply the new migration and load the `ItemStatus` values in `sample_data.json`. Existing `get_item_display_dicts()` test case was updated to test handling of the new field. In the browser, you should be able to:

* Add any number of statuses to an item using checkboxes.
* See statuses (if any) on View Item pages.
* See statuses in the main Search table, and use them to filter data.
* Manage statuses and their display values in the Admin UI.

See screenshots below for how this should look.

Note: I noticed that Title was included as a Basic Field (i.e. not hidden behind the Advanced button) on the View page, but not on Add/Edit. For consistency, I moved it to the Basic section on Add/Edit.

image

image

image

[SYS-1849]: https://uclalibrary.atlassian.net/browse/SYS-1849?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


ftva_lab_data/fixtures/sample_data.json
```diff
@@ -316,5 +316,47 @@
       "general_entry_cataloged_by": "",
       "notes": ""
     }
+  },
```
akohler, line 319: Since item statuses are real data that should always be loaded, please move these into a separate `item_statuses.json` fixture.  Then for the test case, load both fixtures.

ftva_lab_data/views.py
```diff
@@ -111,13 +106,22 @@ def render_search_results_table(request: HttpRequest) -> HttpResponse:
     items = SheetImport.objects.only(*display_fields, "id").order_by("id")
     if search:
         if search_column and search_column in display_fields:
-            # Scoped search to selected column
-            items = items.filter(**{f"{search_column}__icontains": search})
+            # Status is a ManyToMany field, so we need to handle it differently
+            if search_column == "status":
+                # Filter by status using a ManyToMany relationship
+                items = items.filter(status__status__icontains=search)
```
akohler, line 112: When an item has more than one status, the item can appear multiple times in the results - not ideal from the user's POV.  For example, with the sample data as loaded, item 6 has 2 statuses, both containing "Invalid".  A status search for "Invalid" lists the same record twice, once for each matching status:
ftva_django_duplicate_results

I think the solution is to use `distinct()` in the query, as we would via direct SQL:
```
# Current: same record is included twice
>>> SheetImport.objects.filter(status__status__icontains="invalid")
<QuerySet [<SheetImport: id: 6 --- file: Cambodia_Kampuchia --- title: >, <SheetImport: id: 6 --- file: Cambodia_Kampuchia --- title: >]>

# Desired: De-duplication
>>> SheetImport.objects.filter(status__status__icontains="invalid").distinct()
<QuerySet [<SheetImport: id: 6 --- file: Cambodia_Kampuchia --- title: >]>
```

However, see notes in [the documentation](https://docs.djangoproject.com/en/5.2/ref/models/querysets/#distinct) about use of `distinct()` with `order_by()`.  I think we're OK - I got the expected result via a quick manual check, below - but please review and double-check, tweak if needed.
```
>>> from ftva_lab_data.table_config import COLUMNS
>>> display_fields = [field for field, _ in COLUMNS]
>>> items = SheetImport.objects.only(*display_fields, "id").order_by("id")
# Add distinct to filter the set
>>> items = items.filter(status__status__icontains="invalid").distinct()
>>> items
# Just one record, as desired
<QuerySet [<SheetImport: id: 6 --- file: Cambodia_Kampuchia --- title: >]>
```

ftva_lab_data/views.py
```diff
@@ -111,13 +106,22 @@ def render_search_results_table(request: HttpRequest) -> HttpResponse:
...
+                # Scoped search to selected column
+                items = items.filter(**{f"{search_column}__icontains": search})
         else:
             # General CTRL-F-style search across all configured fields
             query = Q()  # start with empty Q() object, always True
             for field in display_fields:  # then add queries for all valid fields
-                query |= Q(**{f"{field}__icontains": search})
+                # Handle status field separately since it is a ManyToMany field
+                if field == "status":
+                    query |= Q(status__status__icontains=search)
```
akohler, line 122: Same de-duplication as in the status-only search will need to be applied here.  Probably the `distinct()` can be applied via the final `filter()`, currently on line 125.

ftva_lab_data/views_utils.py
```diff
@@ -20,6 +20,9 @@ def get_field_value(obj: models.Model, field: str) -> Any:
     Returns the value of the field, or an empty string if the field does not exist
     or is None.
     """
+    # Special case for status field, which is a ManyToMany field
+    if field == "status":
+        return [str(status) for status in obj.status.all()]
```
akohler, line 25: Unless Thelma has said these should be in a specific (non-alphabetical) order, I think these should be alphabetized.  Otherwise, these seem to be ordered by `ItemStatus.id`.

ftva_lab_data/views_utils.py
```diff
@@ -33,8 +36,9 @@ def get_item_display_dicts(item: SheetImport) -> dict[str, Any]:
     """Returns a dictionary of dictionaries. Each top-level dict represents a display section for
     the view_item.html template."""
     header_info = {
-        "File Name": item.file_name,
-        "Title": item.title,
+        "file_name": item.file_name,
+        "title": item.title,
+        "status": [str(status) for status in item.status.all()],
```
akohler, line 41: Same ordering should be applied here as in the general case above.

akohler (changes requested):
@ztucker4 This is a good approach overall.  I like the display of status as "badges", and the slight reorganization of the edit screen to make room for the list of checkboxes is appropriate.  Good decision to move the title field as well.

The current implementation can result in duplicate records in search results, which I think can be fixed pretty easily.  The real status data also needs to be loaded via a separate fixture, and should be ordered in a more meaningful way (unless Thelma wants this specific order).  Once these changes are made, it'll be ready for merge.

Thanks --Andy

ztucker4:
Hi @akohler , this is ready for a second look:
* Statuses are now contained in a separate `item_statuses.json` fixture, used by `tests.py`.
* Statuses are displayed alphabetically on Search and View pages (and also in the Add/Edit form).
* Duplicate results are no longer returned for multiple Status matches. Simply adding `.distinct()` seems to work as desired. I think there's no problem combining this with our use of `order_by()`, since the field we're using for ordering (`SheetImport.id`) is not from a related model.

ftva_lab_data/forms.py
```diff
@@ -62,3 +63,10 @@ class Meta:
             "arsc_lto_tape_id": "ARSC LTO Tape ID",
             "hard_drive_barcode_id": "Hard drive barcode ID",
         }
+
+        widgets = {"status": forms.CheckboxSelectMultiple}
+
+    def __init__(self, *args, **kwargs):
```
akohler, line 69: Instead of forcing the order here in the form, I believe it's better to do it in the model:
https://docs.djangoproject.com/en/5.2/ref/models/options/#ordering

ftva_lab_data/models.py
```diff
@@ -71,3 +74,15 @@ class SheetImport(models.Model):
...
+class ItemStatus(models.Model):
+    """Represents the status of an item in the SheetImport model."""
+
+    status = models.CharField(max_length=50, unique=True)
+
+    def __str__(self):
+        return self.status
+
+    class Meta:
+        verbose_name_plural = "Item statuses"
```
akohler, line 88: Add `ordering = ["status"]` (instead of in the form, mentioned above)
https://docs.djangoproject.com/en/5.2/ref/models/options/#ordering

ftva_lab_data/views_utils.py
```diff
@@ -20,6 +20,10 @@ def get_field_value(obj: models.Model, field: str) -> Any:
     Returns the value of the field, or an empty string if the field does not exist
     or is None.
     """
+    # Special case for status field, which is a ManyToMany field
+    if field == "status":
+        # Order statuses alphabetically
+        return [str(status) for status in obj.status.all().order_by("status")]
```
akohler, line 26: With the addition of `ordering` to `ItemStatus` model, this explicit `order_by` (here and line 42 below) is no longer needed:
```
# I added 3 statuses to this item via UI
>>> item = SheetImport.objects.get(pk=2)
>>> item.status.all()
<QuerySet [<ItemStatus: Duplicated in source data>, <ItemStatus: Invalid inv no>, <ItemStatus: Invalid vault>]>

# No explicit order_by needed now, they're no longer ordered by id
>>> for s in item.status.all():
...   print(s.id, s)
...
2 Duplicated in source data
4 Invalid inv no
3 Invalid vault
```

akohler (changes requested):
@ztucker4 Thanks for making the changes.  I see one more improvement, where the ordering of item status can be moved into the model itself - I had forgotten about that `meta` option.  Once done (and the new migration created & applied), explicit ordering is no longer needed elsewhere unless we want to override what's built into the model.

ztucker4:
Changes made - status ordering is now handled in the model instead of in `forms.py` and `views_utils.py` functions. New migration added, ready for one more review.

PR #15, opened 2025-06-06
Sys 1855/add super editors

ztucker4:
Implements [SYS-1855](https://uclalibrary.atlassian.net/browse/SYS-1855)

Adds a new custom permission (`assign_user`) and corresponding group (`super-editors`) to control access to user assignment. Permission is checked in the `search_results.html` template to conditionally display the "Assign Selected" form, and also checked by the `assign_to_user()` view. 

There are also a few small CSS class changes in the search results template to improve display when the Assign form is not visible.

To test, run the new migration and load the updated `groups_and_permissions.json` fixture. Try logging in with users inside and outside the `super-editors` group. Two new tests are also available in `tests.py`.


akohler (approved):
@ztucker4 This all works as expected: tests pass, and I tested manually as a regular editor, a user in both groups, and a user in neither group.

Users who don't have `super-editors` access might be confused by the "Assign" column and checkboxes which still appear on the search page.  The boxes can still be checked, though the "All" box doesn't toggle all the other boxes for unauthorized users.  However, after looking at the templates, hiding this column would be messy, so I'm OK with leaving as-is for now.  If users are confused or have concerns, we may need to change this.

Merging now, thanks --Andy

PR #22, opened 2025-06-11
SYS-1859:  Fix password change loop

ztucker4:
Fixes [SYS-1859](https://uclalibrary.atlassian.net/browse/SYS-1859)

Adds a "Return to main site" link to the page displayed after a successful password change. This allows users without admin access to easily navigate back to the search results page.

This required a change to the `TEMPLATES["DIRS"]` setting in `settings.py`. This needed to be updated to `[BASE_DIR / "templates"]` to allow partial overrides of templates from other apps (like `django.contrib.admin`). Documentation: https://docs.djangoproject.com/en/5.2/howto/overriding-templates/ 

image

[SYS-1859]: https://uclalibrary.atlassian.net/browse/SYS-1859?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


akohler (approved):
@ztucker4 This is perfect.  I confirmed it works with both the regular `/accounts/password_change/` (from the navigation drop-down) and the admin `/admin/password_change/` (from the Admin toolbar).

Thanks for documenting the use of ` "DIRS": [BASE_DIR / "templates"]`, one TODO off the list!

Merging --Andy

PR #24, opened 2025-06-14
SYS-1864: Add support for full user names

akohler:
Implements [SYS-1864](https://uclalibrary.atlassian.net/browse/SYS-1864).  Version set to `v0.9.1` for deployment.

This PR adds support for displaying and searching full user names, which was a bigger set of changes than I expected:
* Changed the "Assign to..." list of users to display full name instead of username, via built-in `User.get_full_name()`
* Added a calculated property `assigned_user_full_name` to `SheetImport`, to support display of full names for users assigned to those records
* Changed search results template to display the new `assigned_user_full_name`  property
* Changed views and supporting files to search against `assigned_user_full_name`
  * Searches are done against the first name, last name, and username
  * It's not worth the effort to add support for searching full name, in my opinion, but it's possible

Adding this search support led to greater duplication of code.  Searching a single field used the same logic as searching multiple fields, so I refactored that code: a single field is just a list of one field, so I combined the specific and general cases.

The `render_search_results_table` view was doing a little too much, with both the logic for the search and the preparation of data for display.  The search logic is also somewhat complex, and a little fragile, so I wanted it to be testable.  To do this, I pulled the search into a supporting function in `views_utils.get_search_items()`, and wrote enough tests to cover the basic and more complex variations.

Other minor changes:
* I included [SYS-1863](https://uclalibrary.atlassian.net/browse/SYS-1863), moving the Title field from the Basic to Advanced fields for editing, since this was a trivial one-line change.
* Excel files are now ignored for source control.
* I updated `devcontainer.json` to downgrade `reportOptionalMemberAccess` from an error to informational, since the model change references `self.assigned_user.get_full_name()` and pylance can't figure out that `self.assigned_user` is a Django `User` object and has a `get_full_name()` method.  You may need to [rebuild your dev container](https://github.com/UCLALibrary/ftva-lab-data#dev-container) for the change to take effect.

### Testing

No restart needed, as the model changes just add a property (calculated after data is retrieved from the database), so there's no migration from this.  Since ORM queries must run against the database, that's also why I opted not to add support for searching `assigned_user_full_name` itself - it's not in the database.

There are 8 new tests in `SearchTestCase`, bringing the total to 30; all should pass.

[SYS-1864]: https://uclalibrary.atlassian.net/browse/SYS-1864?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[SYS-1863]: https://uclalibrary.atlassian.net/browse/SYS-1863?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -14,8 +14,11 @@
 							"editor.formatOnSave": true,
 							"python.analysis.typeCheckingMode": "standard",
 							"python.analysis.diagnosticSeverityOverrides": {
-								// Disable errors for Django's dynamic "id" model attributes
-								"reportAttributeAccessIssue": "information"
+								// Downgrade errors for Django's dynamic "id" model attributes
+								"reportAttributeAccessIssue": "information",
+								// Downgrade errors when accessing methods on Django's 
+								// related model attributes, like obj.assigned_user.get_full_name()
+								"reportOptionalMemberAccess": "information"
 							},
 							"python.editor.defaultFormatter": "ms-python.black-formatter",
 					"flake8.args": [
--- a/.gitignore
+++ b/.gitignore
@@ -181,3 +181,6 @@ cython_debug/
 
 # Local static files collected when testing gunicorn and whitenoise
 staticfiles/
+
+# Data files
+*.xlsx
--- a/charts/prod-ftvalabdata-values.yaml
+++ b/charts/prod-ftvalabdata-values.yaml
@@ -6,7 +6,7 @@ replicaCount: 1
 
 image:
   repository: uclalibrary/ftva-lab-data
-  tag: v0.9.0
+  tag: v0.9.1
   pullPolicy: Always
 
 nameOverride: ""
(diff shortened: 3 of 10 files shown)


akohler:
@ztucker4 Thanks for the feedback; I was unsure about moving it, but agree that consistency is better.  I've moved Title to the advanced info section on the View template; see http://127.0.0.1:8000/view_item/14/ for an example.

I prefer not to remove the Title header, as it's some useful info at a glance (when it exists).  The main feedback we got was that the fields the editor sees should be the ones which get filled out at a minimum, and since Title as a header is purely informational and space is not an issue on the View template, let's keep it in place up there.

Ready for re-review, thanks --Andy

ztucker4 (approved):
Looks good. Leaving the header title in place is fine with me. Merging now!

PR #25, opened 2025-06-16
SYS 1861: Add command to import status info

henry-r18:
Implements [SYS-1861](https://uclalibrary.atlassian.net/browse/SYS-1861)

Drafting this pull request. Will flesh out later.


henry-r18:
@akohler @ztucker4 

After a lot of experimentation using `pandas`, I reworked `import_status_info.py` to be more straightforward. Rather than handling exceptions from the `SheetImport.objects.get()` method in the `match_record()` function, the script now uses `SheetImport.objects.filter()` and counts the number of items returned.

I was having trouble reproducing Zoe's results of matching all but 94 rows using the `try...except` blocks, but after playing around with the data sets in a Jupyter notebook, I confirmed that using `File Folder Name`, `Sub Folder`, and `File Name` alone uniquely matches Django records for 6460 out of 6741 unique rows with status info from the `Copy of` sheet. As you suggested Andy, I had to fill NA values with empty strings on these columns, and I also stripped them of whitespaces.

I now have the script report out to an Excel spreadsheet the rows from the `Copy of` spreadsheet that resulted in multiple matches and the ones that had no matches, each set on its own sheet. I may need some help figuring out how to proceed with the 205 rows that have multiple matches; I don't get good results when I use the carrier columns from the `Copy of` spreadsheet (`Legacy Carrier Name A` and `Legacy Carrier Name B`). I'm sure I'm missing something there...

That said, I think the script is cleaner now, and 6460 out of 6741 is pretty good.

The command remains the same to run: `python manage.py import_status_info --file_name {PATH_TO_SPREADSHEET}`

akohler:
@henry-r18 Thanks for the updates.  Is this ready for formal review?  If so, I'll turn it into a regular PR and give it a more thorough look.

ftva_lab_data/management/commands/import_status_info.py
```diff
@@ -0,0 +1,139 @@
...
+    status_id_list = []
+    for substring, status_id in status_info_to_model_id_map:
+        if substring.lower() in status_info.lower():
+            status_id_list.append(status_id)
+    # Pandas wants a hashable type to make dropping duplicates possible,
+    # so return a tuple rather than a list
+    return tuple(status_id_list)
+
+
+def match_record(row: pd.Series):
```
akohler, line 43: Please add a return type, `QuerySet[SheetImport]`

ftva_lab_data/tests.py
```diff
@@ -510,3 +513,53 @@ def test_search_finds_carrier_location_in_property_field(self):
             search_fields=["carrier_a_with_location"],
         )
         self.assertEqual(items.all()[0], self.item_basic)
+
+
+class ItemStatusTestCase(TestCase):
+    fixtures = ["sample_data.json", "item_statuses.json"]
```
akohler, line 519: There's currently one test, and it does not use any of the sample records in the `sample_data.json` fixture.  As-is, that fixture doesn't need loading for this test case.

`match_record()` could be tested, but I think there's not much value to doing that.  You could create a `pd.Series` from a list of strings and test that way - no need to create a full dataframe or Excel file - but that's just confirming that database filter queries work, and I think we can trust that.

So, unless you want to add test(s) for `match_records()`, you don't need the `sample_data` fixture.

ftva_lab_data/tests.py
```diff
@@ -510,3 +513,53 @@ def test_search_finds_carrier_location_in_property_field(self):
...
+            ),
+            ("YES: Duplicated in Source Data (invalid inventory_no)", (2, 4)),
+            ("YES:invalid Inventory_no", (4,)),
+            ("Inventory number in filename is incorrect", (1,)),
+            ("", ()),
+        ]
+
+    def test_parse_status_info(self):
+        for status_info, expected_status_ids in self.status_info_test_map:
+            parsed_status = parse_status_info(status_info)
```
akohler, line 562: I think this would be a little better using `subTest` ([documentation](https://docs.python.org/3.13/library/unittest.html#unittest.TestCase.subTest)).  

The benefit is (generally) more specific information if there's a failure.

> When there are very small differences among your tests, for instance some parameters, unittest allows you to distinguish them inside the body of a test method using the [subTest()](https://docs.python.org/3.13/library/unittest.html#unittest.TestCase.subTest) context manager.

For examples in this project, see the first 2 tests in `CleanTapeInfoTestCase`.

henry-r18:
@akohler

I went ahead and changed this to "Ready for review", after updating the README. I'll also push changes to the `process_data.sh` convenience script in a moment, but this should be ready for review. Thanks!

ftva_lab_data/management/commands/import_status_info.py
```diff
@@ -0,0 +1,139 @@
...
+                multiple_matches.append(row)
+
+            if count == 0:
+                no_matches.append(row)
+
+        print(f"{records_updated} records updated")
+        print(f"{len(multiple_matches)} rows returned more than one match")
+        print(f"{len(no_matches)} rows returned no match")
+
+        print("Writing report...")
```
akohler, line 132: Move the Excel report name to a variable and include its name in this message, for convenience.

ftva_lab_data/management/commands/import_status_info.py
```diff
@@ -0,0 +1,139 @@
...
+        tapes_data["status_ids"] = tapes_data["Requires Manual Intervention"].apply(
+            parse_status_info
+        )
+
+        # Fill NA values with empty string, to avoid type issues with Django.
+        # Being explicit with columns, because Pandas complains otherwise.
+        tapes_data.fillna(
+            {"File Folder Name": "", "Sub Folder": "", "File Name": ""}, inplace=True
+        )
+
```
akohler, line 109: Add a status message here, with a count of how many records are left - "Checking {remaining_records_count}..." or similar.  That way we don't have to add up the "updated" and problem counts :)

henry-r18:
I added a command to the `process_data.sh` script that calls this script, **expecting the spreadsheet to be named "copy_dl_sheet_2024-10-18.xlsx" in the working directory**.

akohler (changes requested):
@henry-r18 I think your approach is correct, and that we can move forward with this after a few minor changes noted in separate comments.  Once these small changes are made, this PR can be merged, though the Draft status will need to be removed first.

I ran this using recent data and got results like yours:
```
There are 19641 rows in source data.
Found 180 duplicated rows in source data. Dropping them...
Found 12720 rows without status info in source data. Dropping them...
6459 records updated
205 rows returned more than one match
77 rows returned no match
Writing report...
```
The Excel report has the expected data.

@ztucker4 If this is ready on Friday, can you please give a final review and (if OK) merge?  It doesn't need a deployment chart update, it'll get bundled in with user-facing changes later.

Thanks --Andy

henry-r18:
Thanks @akohler! I'll make these changes on Friday and will request review from Zoe then.

ztucker4 (approved):
This looks good to me! The matching logic is solid, and matching all but 282 rows seems good enough for now. The new test makes sense and passes. README updates are correct, and `process_data.sh` works as expected. Merging!

ztucker4:
Well, apparently not merging, since Andy's change request seems to override my approval. We'll deal with this when Andy's back next week.

PR #26, opened 2025-06-16
SYS-1866: Support carrier search and display

akohler:
Implements [SYS-1866](https://uclalibrary.atlassian.net/browse/SYS-1866).  Sets version to `v0.9.2` for deployment.

This PR adds the ability to search for (tape) carrier information, and adds it to the search results display.  Users can search for either "Carrier A" or "Carrier B"  specifically, and these fields are included in the "All columns" search as well.

Two new columns are added to the search results template (via `partials/search_results_table.html`).  Some other columns were made narrower to make room.

Each "Carrier" column includes data from up to two fields: the carrier field itself, and its corresponding location field.  I chose to implement this via properties in `SheetImport`, for `carrier_a_with_location` and `carrier_b_with_location`.  This allows us to continue using the same generic data retrieval, but I moved that from `views.render_search_results_table()` to `views_utils.get_search_result_data()` for better separation / reuse / documentation.  

### Testing

No restart needed, the model changes do not affect the database.

I chose *not* to add tests for `get_search_result_data()`, as it does nothing complex and the method it uses, `get_field_data()`, already has tests.  However, I did add 2 new simple tests to `SearchTestCase()`, confirming that data in the underlying model fields is retrieved correctly.

There now are 32 tests, all passing.

For manual testing, please just check the search results display to confirm everything fits and looks as good as it can given the data and the display limitations.

### DIFF
--- a/charts/prod-ftvalabdata-values.yaml
+++ b/charts/prod-ftvalabdata-values.yaml
@@ -6,7 +6,7 @@ replicaCount: 1
 
 image:
   repository: uclalibrary/ftva-lab-data
-  tag: v0.9.1
+  tag: v0.9.2
   pullPolicy: Always
 
 nameOverride: ""
--- a/ftva_lab_data/models.py
+++ b/ftva_lab_data/models.py
@@ -79,8 +79,24 @@ class SheetImport(models.Model):
         return f"id: {self.id} --- file: {self.file_name} --- title: {self.title}"
 
     @property
-    def assigned_user_full_name(self):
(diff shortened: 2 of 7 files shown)


henry-r18 (approved):
This looks good to me. Tests pass and behavior is as described.

The table is getting a little cramped on my laptop display, and we may want to consider making the columns resizable and/or show/hide configurable, but that's a separate issue.

I made one minute change to a comment. Approving now.

PR #30, opened 2025-06-20
SYS-1876: Make record_id visible and searchable

ztucker4:
Implements [SYS-1876](https://uclalibrary.atlassian.net/browse/SYS-1876)

Adds `record_id` (The Django internal record identifier) as a searchable data value, and displays it in the following places:
* In search results, sharing a column with the View link
* On View Item pages, with other header metadata
* On Edit Item pages, directly under the page title

Also includes tag bump for deployment. Let me know if you think any of the display/layout could be improved.
image

image

image

### DIFF
--- a/charts/prod-ftvalabdata-values.yaml
+++ b/charts/prod-ftvalabdata-values.yaml
@@ -6,7 +6,7 @@ replicaCount: 1
 
 image:
   repository: uclalibrary/ftva-lab-data
-  tag: v0.9.2
+  tag: v0.9.3
   pullPolicy: Always
 
 nameOverride: ""
--- a/ftva_lab_data/table_config.py
+++ b/ftva_lab_data/table_config.py
@@ -11,4 +11,5 @@ COLUMNS = [
     ("inventory_number", "Inventory number"),
     ("status", "Status"),
     ("assigned_user_full_name", "Assigned user"),
+    ("id", "Record ID"),
 ]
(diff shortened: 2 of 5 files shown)


henry-r18 (commented):
@ztucker4 

I like where you chose to tuck in the Record ID. I tried fitting it as the first column in the table, which I imagine you may have tried as well, and things are just too cramped that way. I think your solution is good.

One issue I'm seeing is with searching on the `Record ID` field. Since our current search mechanism is a sub-string search, I get a lot of pages returned when I search for `1`, for example, or even longer values like `125`.

Is it within the scope of this ticket to make searches on the `Record ID` field exact searches? I imagine a user will have a unique `Record ID`, if they want to pull up a specific record quickly, so I think searches on this field in particular should match exactly.

akohler (commented):
@ztucker4 (and @henry-r18 FYI): The UI changes look great, no changes needed.

After trying it a few times, I decided that record id search should be precise, not a substring search like the others.  For example, short ids like `83` or `351` were basically unfindable.  It doesn't matter if record ids are broadly found during "All columns" search, but a deliberate record id search should be precise.

I modified `views_utils.get_search_result_items()` and added 2 tests to `SearchTestCase`, bringing the total to 35.

Could one of you please review & if these changes look OK, go ahead and merge?

Thanks --Andy

henry-r18:
Both Zoe's initial UI changes and Andy's change to `view_utils` look good with the latest commit. All 35 tests pass. Merging now.

henry-r18 (approved):
Both Zoe's initial UI changes and Andy's change to view_utils look good with the latest commit. All 35 tests pass. Merging now.

PR #32, opened 2025-06-23
SYS-1873: Add visual indicator for required input fields

henry-r18:
Implements [SYS-1873](https://uclalibrary.atlassian.net/browse/SYS-1873)

### Acceptance criteria
- ✅ Update the Add/Edit item display to visually show that Filename is a required field, using an asterisk or other indicator.

### Description
In the Add/Edit item view, labels for required fields now show a red asterisks and a tooltip on hover over the label. The tooltip is in addition to the default browser tooltip that displays on hover over the input element itself.

I replaced the `django-bootstrap` template element with HTML elements to have more control over the label attributes, as I wanted to use the Bootstrap tooltip documented [here](https://getbootstrap.com/docs/5.2/components/tooltips/). The tooltip requires a couple lines of JavaScript, and the asterisks gets appended by a new CSS class. Whether a field is required or not is determined by the `blank=False` parameter of the underlying model field.

So far, `file_name` is the only field with `blank=False` (i.e. cannot be blank, i.e. is required). Any additional fields set to `blank=False` should get the asterisks in the label automatically now.

### Testing
I added one unit test, which checks that the label for the file name field in the `add_item` view has `required-field-label` in its class list. This class triggers the CSS `::after` pseudo-element to append an asterisks. Since it's a pseudo-element, it doesn't show up directly in the HTML returned to the test `GET` call, but I think we can trust the CSS will work.

Since `add_item` and `edit_item` share a template, I figured it was okay to just test `add_item`. Let me know if you disagree though, and I'm happy to add to the test.

One manual test you can do to confirm the changes work as intended is to change any number of `SheetImport` fields in `models.py` to have `blank=False` as a parameter. Those fields should then show up with an asterisk and tooltip.

### Reference screen capture
screencap

[SYS-1873]: https://uclalibrary.atlassian.net/browse/SYS-1873?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/charts/prod-ftvalabdata-values.yaml
+++ b/charts/prod-ftvalabdata-values.yaml
@@ -6,7 +6,7 @@ replicaCount: 1
 
 image:
   repository: uclalibrary/ftva-lab-data
-  tag: v0.9.3
+  tag: v0.9.4
   pullPolicy: Always
 
 nameOverride: ""
--- a/ftva_lab_data/templates/add_edit_item.html
+++ b/ftva_lab_data/templates/add_edit_item.html
@@ -33,7 +33,20 @@
             <div class="col">
             {% for field in form %}
                 {% if field.name in basic_fields %}
-                    {% bootstrap_field field %}
+                    <!-- `field.field` accesses actual field instance, rather than wrapper -->
+                    {% if field.field.required %}
+                        <div class="mb-3">
+                            <label
+                                for="id_file_name"
+                                class="form-label required-field-label"
+                                data-bs-toggle="tooltip"
+                                title="Required field"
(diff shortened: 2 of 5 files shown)


ztucker4 (approved):
This all looks good to me! I like the approach of checking `field.field.required` in the template so that if we add more `blank=False` fields in the model they will automatically get the CSS and tooltip. The display itself looks nice and clearly shows required-ness. Merging!

PR #33, opened 2025-06-24
SYS-1874: Improve pagination controls

ztucker4:
Implements [SYS-1874](https://uclalibrary.atlassian.net/browse/SYS-1874)

Uses Django's "elided page" feature to display a small range of page numbers for navigation in the search results table. Both top and bottom pagination controls should now display direct links to the two pages before and after the current page, plus links to the first and last pages and the pre-existing "Previous" and "Next" buttons.

The number of pages displayed in the controls is easily adjusted with the `on_each_side` and `on_ends` parameters of `get_elided_page_range()`, so let me know if these should be changed.

image

[SYS-1874]: https://uclalibrary.atlassian.net/browse/SYS-1874?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


ftva_lab_data/templates/partials/search_results_table.html
```diff
@@ -92,16 +109,33 @@
...
-    <button class="btn btn-primary" {% if page_obj.has_previous %}
-        hx-get="{% url 'render_table' %}?page={{ page_obj.previous_page_number }}&search={{ search }}&search_column={{ search_column }}"
-        hx-target="#table-container" {% else %} disabled {% endif %}>Previous</button>
-
-    <span>Page {{ page_obj.number }} of {{ page_obj.paginator.num_pages }}</span>
-
-    <button class="btn btn-primary" {% if page_obj.has_next %}
-        hx-get="{% url 'render_table' %}?page={{ page_obj.next_page_number }}&search={{ search }}&search_column={{ search_column }}"
-        hx-target="#table-container" {% else %} disabled {% endif %}>Next</button>
+    <button class="btn btn-primary"
```
akohler, line 112: This page navigation code seems to be exactly the same here and above at lines 1-29.  Can it be moved to another partial template and included in both places, instead of explicit repetition?

ftva_lab_data/views.py
```diff
@@ -157,6 +157,12 @@ def render_search_results_table(request: HttpRequest) -> HttpResponse:
 
     paginator = Paginator(items, 10)
     page_obj = paginator.get_page(page)
+    # Convert elided page range to list to allow multiple iterations in template
+    elided_page_range = list(
+        paginator.get_elided_page_range(
+            number=page_obj.number, on_each_side=2, on_ends=1
```
akohler, line 163: I think we should offer more pages `on_each_side`, maybe 4 or even 5, to allow for quicker navigation through large result sets.  I agree with `on_ends=1`, I don't see value to expanding that.

akohler (changes requested):
@ztucker4 I like this a lot.  I think expanding the number of pages shown will be helpful, so the user can jump by more than the current 2 pages at a time.  I'd also like the duplicate navigation template code to be pulled into its own file for easier maintenance.
Thanks --Andy

ztucker4:
@akohler, this is ready for another look. The pagination controls have been moved into a new partial template, and the number of pages per side has been increased to 5 (which looks reasonable to me even with large page numbers at small screen sizes).

akohler (approved):
@ztucker4 Thanks for making the changes.  The new partial template helps understand and maintain the code, and the use of 5 pages on each side is a good balance for navigating result sets of various sizes.  Merging now, thanks.

PR #34, opened 2025-06-26
SYS-1880: Export search results

ztucker4:
Implements [SYS-1880](https://uclalibrary.atlassian.net/browse/SYS-1880)

Adds an "Export to Excel" button, visible to all signed-in users, that will export the current set of search results to an Excel file called `FTVA_DL_search_results_{timestamp}.xlsx`. The export file contains all fields associated with each `SheetImport` object, not just those visible on the search results screen.

Includes minor refactoring of `main.js` to re-use the `setOrUpdate()` function written for user assignment, and updated docstrings throughout `views.py` to use the new Sphinx markup style.

This PR adds `xlsxwriter` as a new dependency in `requirements.txt`. Rebuild the Django container to install this.

image

[SYS-1880]: https://uclalibrary.atlassian.net/browse/SYS-1880?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/ftva_lab_data/tests.py
+++ b/ftva_lab_data/tests.py
@@ -729,12 +729,6 @@ class DataExportTestCase(TestCase):
         data_dicts = [row.__dict__ for row in self.rows]
         export_data = format_data_for_export(data_dicts)
 
-        # Check that the carrier_a_location column is next to the carrier_a column
-        self.assertIn("carrier_a", export_data.columns)
-        self.assertIn("carrier_a_location", export_data.columns)
-        carrier_a_index = export_data.columns.get_loc("carrier_a")
-        carrier_a_location_index = export_data.columns.get_loc("carrier_a_location")
-        self.assertEqual(carrier_a_index + 1, carrier_a_location_index)
         # Test that the assigned user's full name is formatted correctly
         self.assertIn("Example User", export_data["assigned_user"].values)
         # Check that the Statuses are concatenated correctly
--- a/ftva_lab_data/tests.py
+++ b/ftva_lab_data/tests.py
@@ -22,6 +22,7 @@ from ftva_lab_data.views_utils import (
     get_item_display_dicts,
     get_search_result_items,
     get_items_per_page_options,
+    format_data_for_export,
 )
 from ftva_lab_data.table_config import COLUMNS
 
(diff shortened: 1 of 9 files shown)


akohler (changes requested):
@ztucker4 This works very well.  I tried with all records, and with smaller search result sets, and files were created / named / downloaded, using my browser settings.

See detailed comment on the Excel -> Response changes to (I hope) solve a type error - not a functionality problem but I think this code improvement is worth doing.  Also some minor docstring formatting changes seem necessary for VSCode to properly parse & display the extra details.

For the Excel output itself:
* Columns are ordered the same as in the model (I think).  When I added the carrier location fields, I put them at the end of the model.  Can you please update the model to move them, so `carrier_a`, `carrier_a_location`, etc?
* `assigned_user_id` is just that, the numerical user_id.  Can this be the `assigned_user_full_name` property instead?
* `status` is not included, probably because of its many-to-many relation.  What would it take to include this?  If the `assigned_user_full_name` property can be included, probably status could also by creating a property on `SheetImport`, something like `status_list`, returning a concatenated list of statuses.

Thanks --Andy

akohler:
I agree with Henry's suggestion for showing something in the UI when a download is being generated.  Something specific is probably better than just disabling the button after the user starts a download.

For the position, either is OK with me. Associating it with the search results via the navigation controls makes sense... but this page always shows search results, and the user acts on them via filtering, which is the top row of controls.  @ztucker4 I defer to you on this decision, as you are more UX-aware than I am :punt emoji:

ftva_lab_data/views_utils.py
```diff
@@ -214,3 +238,52 @@ def get_items_per_page_options() -> list[int]:
...
+            data_dict["assigned_user"] = assigned_user_full_name
+        else:
+            data_dict["assigned_user"] = ""
+        # Remove the '_state' field added by Django
+        data_dict.pop("_state", None)
+
+    # Convert rows to DataFrame for exporting
+    df = pd.DataFrame(data_dicts)
+
+    # Reorder carrier_a_location and carrier_b_location to be next to carrier_a and carrier_b
```
akohler, line 272: This code is not needed, if you change the order of the fields in the model.  I checked by disabling this code and moving the location fields, like this:
```
    carrier_a = models.CharField(max_length=100, blank=True)
    carrier_a_location = models.CharField(max_length=25, blank=True)
    carrier_b = models.CharField(max_length=100, blank=True)
    carrier_b_location = models.CharField(max_length=25, blank=True)
```
Removing the carrier reordering code also will resolve several "Operator not supported" type-related errors pylance is showing with these lines of code.

ftva_lab_data/tests.py
```diff
@@ -704,3 +705,40 @@ def test_having_required_field_allows_save(self):
...
+        SheetImport.objects.get(pk=3).status.add(1, 2)
+
+        # Create a list of SheetImport objects to test with
+        self.rows = [SheetImport.objects.get(pk=2), SheetImport.objects.get(pk=3)]
+
+    def test_format_data_for_export(self):
+        data_dicts = [row.__dict__ for row in self.rows]
+        export_data = format_data_for_export(data_dicts)
+
+        # Check that the carrier_a_location column is next to the carrier_a column
```
akohler, line 732: With the carrier location fields moved in the model, that part of this test could be removed (though it does still pass... but by changing the model, testing position for carrier is not necessary).

static/js/main.js
```diff
@@ -59,23 +71,58 @@ document.addEventListener("DOMContentLoaded", function () {
...
+  const exportForm = document.getElementById("export-form");
+  const spinner = document.getElementById("export-spinner");
+  if (exportBtn && exportForm && spinner) {
+    exportBtn.addEventListener("click", function () {
+      spinner.style.display = "block";
+      exportForm.submit();
+      // Hide spinner after a constant delay
+      setTimeout(() => {
+      spinner.style.display = "none";
+      }, 60000);
```
akohler, line 125: I don't like the long-running spinner.  Unless the user notices their browser has finished the download, they may sit and wait until the spinner expires.  The download does not prevent them from doing more with the application while the download is actually happening, or when the download finishes and the spinner is still running.

I expect downloads will be used mostly for specific sets rather than all rows.  Changing to a 10-15 second spinner seems like a better user experience, one we can revisit if users report problems.

akohler (changes requested):
@ztucker4 This is almost ready.  The output Excel looks great, with full user names and all assigned statuses, and carrier location fields in the right place (though see my detailed comments about just doing the carrier locations in the model instead).  Thanks also for the docstring updates.

I think the spinner timing needs to be cut way down, to 10-15 seconds at most.  It's not ideal either way, but I think most exports will be smaller.  Let's try that, and see if users have feedback for us once they start using it.

Thanks --Andy

ztucker4:
Ready for a hopefully final look. I reordered the carrier location fields in the model, which definitely looks cleaner. The spinner should now timeout after 10 seconds, which was close to the time required to export the ~6500 rows with Status data.

henry-r18:
@ztucker4 As Andy said, the Excel export looks great, and all functionality is as described.

I see why HTMX won't work and that there's no easy way to tell when the download is complete. One additional approach that may be worth considering later on would be to prevent the default form submission and use asynchronous JavaScript to get the spreadsheet blob, then trigger a download of the blob from JavaScript.

To be clear, I don't think it's worth doing now. Especially with the reduced spinner timeout, this looks great to me, though I'll let Andy sign off.

akohler (approved):
@ztucker4 I like the shorter timer, and get similar results with the 6600 status rows export in about 10 seconds.  If this functionality causes confusion later, we'll revisit it then.

The code is simpler, pylance is happy, everything works as it should - merging, thanks!

PR #35, opened 2025-06-26
SYS-1879: add records per page control

henry-r18:
Implements [SYS-1879](https://uclalibrary.atlassian.net/browse/SYS-1879)

### Acceptance criteria
- ✅ User can set number of search result records per page
- ✅ This should persist across searches, so user has to set it only once per session.
    - ❌ It does not need to persist across sessions, via user-specific settings.

### Details
I changed `partials\pagination_controls.html` into a row with three columns, then added the controls in the left-most column of the row. The user select from a drop-down of per-page options. For now, this list is hard-coded in `views.py`.

The per-page selection should persist when the user refreshes the page or navigates away from the table to the `add_item` or `edit_item` routes. As the user pages through the table, the setting should persist, as it should when a search term is entered.

One thing to note is that when the per-page selection changes, the pagination recalculates, and so the table resets to page 1. This is similar to the behavior when the search inputs change.

### Testing
I added one test to the `TablePaginationTestCase`. The test demonstrates that given various request parameters, the `per_page` value on the `Paginator` object, from which the table gets its pages, and the `items_per_page` value in the session store stay in sync. I could add a test to actually validate the count of `<tr>` sub-strings in the actual response content is as expected; let me know if you think that's worthwhile.

screencap

[SYS-1879]: https://uclalibrary.atlassian.net/browse/SYS-1879?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/ftva_lab_data/templates/partials/pagination_controls.html
+++ b/ftva_lab_data/templates/partials/pagination_controls.html
@@ -1,11 +1,16 @@
 <div class="row">
+    {% comment %}This conditional controls whether the two <div> elements
+    on either side of the main pagination controls are shown or not,
+    as this partial is used both above and below the table,
+    and we only want to show the additional controls above the table.
+    {% endcomment %}
     {% if show_auxiliary_controls %}
         <div class="col-2 d-flex align-items-center gap-2">
             <select
                 name="items_per_page"
                 class="form-select form-select-md"
                 style="width: fit-content;"
-                hx-get="{% url 'render_table' %}"
+                hx-get="{% url 'render_table' %}?search={{ search }}&search_column={{ search_column }}"
                 hx-target="#table-container"    
             >
                 {% for option in items_per_page_options %}
@@ -46,7 +51,8 @@
             Next
         </button>
     </div>
-
+    
+    {% comment %}See comment above regarding this conditional.{% endcomment %}
     {% if show_auxiliary_controls %}
         <div class="col-2">
             <!-- Added to keep row balanced and main pagination controls centered.
--- a/ftva_lab_data/templates/partials/search_results_table.html
+++ b/ftva_lab_data/templates/partials/search_results_table.html
@@ -1,3 +1,7 @@
+{% comment %}The `show_auxiliary_controls` attribute controls whether or not
+to render <div> elements for controls on either side of the main pagination controls.
+See comment on partial as well.
+{% endcomment %}
 {% include "partials/pagination_controls.html" with show_auxiliary_controls=True %}
 
 <table class="table table-hover" style="table-layout: fixed;">
@@ -79,7 +83,8 @@
     </tbody>
 </table>
 
-{% include "partials/pagination_controls.html" %}
+{% comment %}See comment above regarding `show_auxiliary_controls` attribute.{% endcomment %}
+{% include "partials/pagination_controls.html" with show_auxiliary_controls=False %}
 
 <!-- Hidden input to store the current page number for HTMX requests -->
 <!-- Used to maintain state after POST requests to assign items to users -->
--- a/ftva_lab_data/tests.py
+++ b/ftva_lab_data/tests.py
@@ -20,6 +20,7 @@ from ftva_lab_data.views_utils import (
     get_field_value,
     get_item_display_dicts,
     get_search_result_items,
+    get_items_per_page_options,
 )
 from ftva_lab_data.table_config import COLUMNS
 
@@ -201,7 +202,7 @@ class TablePaginationTestCase(TestCase):
         SheetImport.objects.all().delete()
 
         # Create enough SheetImport objects to require pagination
-        # (i.e. more than the default page size of 10)
+        # (i.e. more than the maximum page size allowed by `items_per_page`)
         for i in range(150):
             SheetImport.objects.create(file_name=f"test_file_{i}", id=i)
         # Create SheetImport objects with different filenames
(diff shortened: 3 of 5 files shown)


ftva_lab_data/templates/partials/pagination_controls.html
```diff
@@ -1,29 +1,56 @@
...
-    {% endfor %}
-    <button class="btn btn-primary"
-        {% if page_obj.has_next %}
-            hx-get="{% url 'render_table' %}?page={{ page_obj.next_page_number }}&search={{ search }}&search_column={{ search_column }}"
-            hx-target="#table-container"
-        {% else %} disabled {% endif %}>
-        Next
-    </button>
+<div class="row">
+    {% if show_auxiliary_controls %}
```
akohler, line 2: Please add a comment clarifying the purpose of this check.  It took a few looks for me to understand that this applies only to the top row of nav controls.

ftva_lab_data/templates/partials/search_results_table.html
```diff
@@ -1,4 +1,4 @@
-{% include "partials/pagination_controls.html" %}
+{% include "partials/pagination_controls.html" with show_auxiliary_controls=True %}
```
akohler, line 1: It took a few looks for me to understand that this applies only to the top row of nav controls.  I think this line is OK, but the second call (line 82) to show the bottom nav controls would be better with an explicit `with show_auxiliary_controls=False`, to make the intent clear.

ftva_lab_data/tests.py
```diff
@@ -202,11 +202,11 @@ def setUp(self):
 
         # Create enough SheetImport objects to require pagination
         # (i.e. more than the default page size of 10)
```
akohler, line 204: The comment should be updated to reflect this is now to test pagination with more than the max value of `items_per_page`.

ftva_lab_data/tests.py
```diff
@@ -239,6 +239,26 @@ def test_search_results_with_pagination(self):
         self.assertIn("test_file_10", page_2_content)
         self.assertNotIn("unique_file_1", page_2_content)
 
+    def test_items_per_page_control(self):
+        url = reverse("render_table")
+        test_values = ["", 10, 20, 50, 100, "foobar"]
```
akohler, line 244: The 2 string values could use some clarification.  From reading the code in the view, I understand that if the value can't be cast to `int`, the default value will be used and the exception swallowed.  But there's nothing in this test related to that, so the test is not intuitive.

In fact, I think it would be better to add a test explicitly for non-int values - one should be enough, since `int("")` and `int("foobar")` both give a `ValueError`.  Then the test can assert that `pagination_per_page` is the default value.... which might lead to a little refactor to move the options into a dedicated function, so they're easier to include in tests or other reuse.

ftva_lab_data/views.py
```diff
@@ -147,15 +147,32 @@ def render_search_results_table(request: HttpRequest) -> HttpResponse:
...
 
+    display_fields = [field for field, _ in COLUMNS]
     # If there's a specific search column, use it;
     # otherwise, search in all display fields.
     search_fields = [search_column] if search_column else display_fields
 
     items = get_search_result_items(search, search_fields)
 
-    paginator = Paginator(items, 10)
+    items_per_page_options = [10, 20, 50, 100]  # hard-coded here for now
```
akohler, line 159: See my comment on tests. It's probably worth moving this 1 line to a trivial `view_utils.get_items_per_page_options()`, so it can be included in tests and you can easily determine the default value from the list (assuming it's always intended to be the first in the list... might be better to refactor that too).

akohler (changes requested):
@henry-r18 I added several comments about the code, all for minor refactoring.  Users may have feedback about the placement of the "Records per page" control: I didn't see it on the far left at first, expecting it closer to the page navigation controls, but it's OK where it is for now.

I understand that changing the items/page will have an effect on the results, as the Paginator takes `per_page` via initializer and doesn't seem to have a [built-in way](https://github.com/django/django/blob/stable/5.2.x/django/core/paginator.py) to change `per_page` in an existing Paginator... though I wonder if just setting `paginator.per_page = new_value` would work?

However, the UX needs to be improved, so that the user does not lose their current search when `per_page` is changed after doing an explicit search.  Currently, the search index and the search term remain unchanged, which definitely makes the user wonder why their results do not reflect the info the UI shows after updating `per_page`.

The current search index and terms should continue to be kept as the UI shows, but a search must be (re)done when `per_page` changes so the user is not forced to redo their search.  Currently, even clicking in the search term box and hitting Enter doesn't redo the search, the user has to edit the search term to trigger an update.

@ztucker4 made [changes to javascript](https://github.com/UCLALibrary/ftva-lab-data/pull/34/files#diff-431ce3d3e025bfd0726c39bbfa7bea8c5b29e7b1df9e8a8b19ed4769d98c5d61) in her PR #34 which may support that persistence, so please take a look at that.

If it's not possible to redo the current search on `per_page` change, the search UI probably needs to be reset to make it clear that the old search is gone.  But I think that should not be needed.

Thanks --Andy

ftva_lab_data/templates/partials/pagination_controls.html
henry-r18: I chose to add comments using Django's template syntax, so the comments wouldn't show in the HTML in the browser developer tools, but I can use HTML comments for consistency, if you'd prefer.

ftva_lab_data/tests.py
henry-r18: I added a new test to address invalid options. I see what you mean about only needing one string value, so I kept just the empty string. Then it occurred to me that a Boolean might cause an issue, so I added a Boolean as an invalid test value. Still works, as `int()` still throws a `ValueError`.

ftva_lab_data/templates/partials/pagination_controls.html
akohler: Django template comments are appropriate, thanks.

akohler (approved):
@henry-r18 This all looks great now, thanks for making the changes.  The search index and terms now persist and the search is refreshed correctly when the user changes the `per_page` setting.  Code comments and tests are clear and have the necessary context now.  Merging!

@ztucker4 FYI

PR #40, opened 2025-06-30
SYS-1883: Update status for records with inline notes

henry-r18:
Implements [SYS-1883](https://uclalibrary.atlassian.net/browse/SYS-1883)

### Acceptance criteria

- ✅ Records with inline notes in the path fields are labelled with the “Needs review” status

### Discussion
This pull request adds a function to the `clean_imported_data` management command which filters the `SheetImport` objects for records where the three path-like fields (`file_folder_name`, `sub_folder_name`, and `file_name`) have sub-strings between square brackets. Per discussion on Slack, we decided such sub-strings are most likely inline notes. The function then applies the new "Needs review" status to matching records.

I also went ahead and updated docstrings to our new preferred format in this management command.

### Testing
I didn't create any unit tests; let me know if you think a unit test for the regex is necessary. The count of records updated may vary between versions of the DL Sheet, but should be in the low-hundreds.

The command can be run with `python manage.py clean_imported_data`, and it will also run as part of `process_data.sh`.

[SYS-1883]: https://uclalibrary.atlassian.net/browse/SYS-1883?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/ftva_lab_data/management/commands/clean_imported_data.py
+++ b/ftva_lab_data/management/commands/clean_imported_data.py
@@ -1,6 +1,7 @@
 import re
 from django.core.management.base import BaseCommand
-from ftva_lab_data.models import SheetImport
+from ftva_lab_data.models import SheetImport, ItemStatus
+from django.db.models import Q
 
 
 def _is_header_record(record: SheetImport) -> bool:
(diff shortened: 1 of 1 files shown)


henry-r18:
Thanks for the note regarding object references--very handy!

PR #44, opened 2025-07-02
SYS-1892: Sync button spinner with request

henry-r18:
### Acceptance criteria
- [ ] Spinner is applied when user clicks export button and request is issued, then removed when download of exported data completes

### Description
This PR modifies the export button so that it no longer triggers a form submit, but rather triggers a JavaScript function via HTMX on click. The JavaScript changes consolidate the existing functionality into a single asynchronous function, which handles the click event from the export button, applies styling, and issues a request against the export endpoint.

The JavaScript the takes the data returned from the Django view, encodes it in a URL via [creatObjectURL()](https://developer.mozilla.org/en-US/docs/Web/API/URL/createObjectURL_static), then applies the URL to a temporary anchor tag. The download action on the anchor tag is then triggered, and styling on the button is returned to normal.

The function also creates a `FormData` object from the existing table filters form, to alleviate the need for the functions that kept the hidden input fields in sync.

### Testing
I didn't add any automated tests, as there are no changes to Python. For manual testing, please ensure the download works with filters applied, and also with no filters applied, and that the button behaves as described.

screencap

### DIFF
--- a/ftva_lab_data/templates/search_results.html
+++ b/ftva_lab_data/templates/search_results.html
@@ -14,7 +14,7 @@
     <!-- HTMX will listen for triggers from enclosed controls,
     then issue GET requests to 'render_table' URL
     and replace content of #table-container with result -->
-    <form class="d-flex gap-2 align-items-stretch w-50" hx-get="{% url 'render_table' %}" hx-target="#table-container"
+    <form id="table-filters-form" class="d-flex gap-2 align-items-stretch w-50" hx-get="{% url 'render_table' %}" hx-target="#table-container"
         hx-trigger="change from:find select, keyup changed delay:300ms from:find input, clear from:find input">
         <select name="search_column" class="form-select form-select-md">
             <option value="">All columns</option>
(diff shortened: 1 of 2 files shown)


akohler (commented):
@henry-r18 This is great! I tested with small (60) / medium (6000) / full (26K) records, and the spinner was closely synchronized to my browser download icon each time.  Tested with both Firefox and Edge, on Windows.  I assume you tested on Mac?  If not, that should be checked.

I understand how it's hooked into the Django template, and I can follow the flow of the Javascript itself, though I'm not familiar with some of the specifics.  Is a temporary anchor tag used because that natively supports the idea of "clicking a URL", so it's the best fit even though that all takes place behind the scenes?

@ztucker4 I'd like you to give the final review / feedback.  If all is OK with you, go ahead and approve & merge.  I'm not requesting any changes, though I suggested a change to the default filename setting.

This may also need changes if the download performance work conflicts with this.  I think that as long as the `export_search_results` view still returns an `HttpResponse` with a filename in the disposition, all will be fine.  But if you are moving towards a streaming response, that may well be a significant difference from this spinner work that we'll need to discuss in more depth.

Thanks --Andy

ztucker4:
@henry-r18 This is awesome! I tested in Firefox, Chrome, and Safari on Mac. The spinner appears and disappears in time with the file request and response. 

This also appears to already be compatible with all of my performance work. If the view returns a `StreamingHttpResponse` instead of an `HttpResponse`, the button still works as expected.

I agree with Andy's comment about the default filename - if we have one, it should be `FTVA_DL_search_results.xlsx` (which I'll likely update to `FTVA_DL_search_results.csv` after performance updates.

akohler (approved):
@henry-r18 I think this is fine. I triggered an exception in the view via the standard `i = 1/0` and got a pop-up alerting me 
```
Error: An error occured while exporting data.
Internal Server Error (500)
```
@ztucker4 If all is OK with you, please approve & merge, then tweak as needed for your performance ticket.
Thanks --Andy

ztucker4 (approved):
This also looks good to me. I confirmed that the new error alert works as expected, and the spinner looks great.
Merging now, thanks!

PR #47, opened 2025-07-08
SYS-1895: Fix over-broad file folder naming

akohler:
Fixes [SYS-1895](https://uclalibrary.atlassian.net/browse/SYS-1895).  Not tagged for deployment as this is an internal script.

This PR fixes a bug in `clean_imported_data.set_file_folder_names()`.  This method fills in the `file_folder_name` field by copying the value from a previous row, where safe.  The previous version was documented as doing this only for rows which have a value in the `file_name` field... but it did not actually check that, so some records were getting updated when they should not be.

The method now only updates records which have other file information (subfolder name and/or file name).  A new utility function `_has_file_info(record)`  supports this, by testing whether the given record has any of `file_folder_name`, `sub_folder_name`, or `file_name` set.  This is broader than needed, but the updated logic in `set_file_folder_names()` ensures this is only checked when `file_folder_name` of the current record is not set.

### Testing

No new tests; the existing one still passes.  I did add a record to the test fixture which fits the new logic (no file info, but it has other data so not a hard-drive only record or empty record or header record).  I updated the comments in that test case to reflect the exclusion of the new test record.

Before & after runs with the same source data files:
* Before: Folder names set: 12372
* After: Folder names set: 12315

[SYS-1895]: https://uclalibrary.atlassian.net/browse/SYS-1895?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


henry-r18 (approved):
@akohler All tests still pass, and I get a similar count of folder names set, using an older `ftva_dl_sheet.xlsx` as input. All good! Merging now.

FYI// @ztucker4

PR #48, opened 2025-07-10
SYS-1901: Add search fields for `SheetImportAdmin` class

henry-r18:
Implements [SYS-1901](https://uclalibrary.atlassian.net/browse/SYS-1901)

### Acceptance criteria
- [x] When authorized users click on [Sheet imports in the Admin interface](http://127.0.0.1:8000/admin/ftva_lab_data/sheetimport/), they have the option to enter a specific record id to go quickly to that specific record.  This is needed so they can view the History for that record.

### Description
I added a new class to `ftva_lab_data/admin.py`, which extends the `SimpleHistoryAdmin` class in use on the admin site for `SheetImport`. I also changed the `admin.site.register` lines to `@admin.register`, for consistency with how it was done in our [TERRA](https://github.com/UCLALibrary/terra/blob/main/terra/admin.py#L67) application.

I don't think the `ItemStatusAdmin` class was doing anything prior to this change, as the class was declared but not registered on the admin site for the `ItemStatus` model. I went ahead and registered it, so item statuses are now also searchable on the admin site, though that's probably not necessary.

### Testing
No automated tests, as a unit test would just be demonstrating that Django works, I think. Please manually confirm that the sheet import records are searchable by ID on the admin site linked above.

[SYS-1901]: https://uclalibrary.atlassian.net/browse/SYS-1901?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/ftva_lab_data/admin.py
+++ b/ftva_lab_data/admin.py
@@ -11,4 +11,4 @@ class ItemStatusAdmin(admin.ModelAdmin):
 
 @admin.register(SheetImport)
 class SheetImportAdmin(SimpleHistoryAdmin):
-    search_fields = ("id",)
+    search_fields = ("id__exact",)


ftva_lab_data/admin.py
```diff
@@ -2,10 +2,13 @@
...
 class ItemStatusAdmin(admin.ModelAdmin):
-    list_display = ("status", "display_name")
-    search_fields = ("status", "display_name")
+    list_display = ("status",)
+    search_fields = ("status",)
+
+
+@admin.register(SheetImport)
+class SheetImportAdmin(SimpleHistoryAdmin):
+    search_fields = ("id",)
```
akohler, line 14: I think this needs to be `id__exact`.  Currently, it's doing an `icontains` search, so when I search for shorter id values like `3` I get thousands of records.  We need to be able to go to the specific id requested.
https://docs.djangoproject.com/en/5.2/ref/contrib/admin/#django.contrib.admin.ModelAdmin.search_fields

henry-r18:
@akohler Made the ID search exact--thanks for catching that!

PR #54, opened 2025-07-25
SYS-1929: Encode search to make safe for URL query strings

henry-r18:
Fixes [SYS-1929](https://uclalibrary.atlassian.net/browse/SYS-1929)

### Acceptance criteria
- [x] Reserved characters in search strings are properly encoded in URL query strings
- [x] When navigating from item view or other pages to the home table view, search is preserved, so that the correct number of results display

### Description
This PR updates `views.py` to return an encoded version of the string entered in the search input, safe for use in URL query strings. This `url_encoded_search` is then passed to the relevant templates in the view contexts, and from there passed between pages in the URL query strings to address the bug caused by reserved characters (e.g. `&`, `+`, `#`, etc) in the search param.

### Testing
No new automated tests. All 44 existing tests pass. Please manually confirm all the example searches described in the Jira issue yield the expected number of results, particularly when navigating from the item view to the table view via the "Back to Search" button.

### Discussion
**TL;DR**
HTMX offers a lot of convenience via attributes like [hx-include](https://htmx.org/attributes/hx-include/), but adopting it more fully would push the application toward a single-page application (SPA) style. That would likely require a fair amount of refactoring, and it ain't broken, so perhaps it's wise to not fix it.

--

Addressing this bug made me realize that there are likely ways to more fully utilize HTMX. Currently, we employ a hybrid approach between idiomatic Django and HTMX. Consider the following code snippet:

```
<div id="table-container" class="table-responsive container-fluid p-4"
     hx-get="{% url 'render_table' %}?{{ url_encoded_search }}&search_column={{ search_column }}&page={{ page }}"
     hx-trigger="load">
</div>
```

Here we use the `hx-get` attribute to make a request to the `render_table` view, and we pass parameters in the URL query string (which was just updated in this PR to encode the `search` parameter). For the `search` and `search_column` params, we could use [hx-include](https://htmx.org/attributes/hx-include/), which gets values from child `input` and `select` elements of the target DOM element, then automatically includes them as encoded params in the prepared AJAX request.

In `partials/search_results_table.html`, we include an anchor tag for each row in the table, which navigates to the item view for the associated record:

```
<a class="btn btn-link btn-sm p-0"
href="{% url 'view_item' row.id %}?{{ url_encoded_search }}&search_column={{ search_column }}&page={{ page_obj.number }}">View</a> 
```

If this element were to use `hx-get` rather than `href`, we could utilize HTMX to swap in the resulting item view, rather than initiating a full page reload. A similar pattern could be followed for other navigation that results in a full page reload. This would make the front-end of the application more of a single-page application (SPA).

Here are some pros and cons for utilizing HTMX more fully, as I see them:
**Pros**
HTMX provides a lot of convenience for doing things like encoding URL params. As it relates to the bug at hand, HTMX could allow us to avoid long query strings in links by using the `hx-include` attribute to refer to the `#table-filters-form` element as the source for the search parameters.

**Cons**
The more we adopt HTMX, the more "locked in" to it we get. It would push the application more toward a SPA style, which could introduce more awkwardness in working with Django's built-in features.

[SYS-1929]: https://uclalibrary.atlassian.net/browse/SYS-1929?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/ftva_lab_data/templates/partials/pagination_controls.html
+++ b/ftva_lab_data/templates/partials/pagination_controls.html
@@ -10,7 +10,8 @@
                 name="items_per_page"
                 class="form-select form-select-md"
                 style="width: fit-content;"
-                hx-get="{% url 'render_table' %}?{{ url_encoded_search }}&search_column={{ search_column }}"
+                hx-get="{% url 'render_table' %}"
+                hx-include="#table-filters-form"
                 hx-target="#table-container"    
             >
                 {% for option in items_per_page_options %}
@@ -25,7 +26,8 @@
     <div class="col d-flex justify-content-center align-items-center gap-2">
         <button class="btn btn-primary"
             {% if page_obj.has_previous %}
-                hx-get="{% url 'render_table' %}?page={{ page_obj.previous_page_number }}&{{ url_encoded_search }}&search_column={{ search_column }}"
+                hx-get="{% url 'render_table' %}?page={{ page_obj.previous_page_number }}"
+                hx-include="#table-filters-form"
                 hx-target="#table-container"
             {% else %} disabled {% endif %}>
             Previous
@@ -37,7 +39,8 @@
                 <span class="btn btn-secondary active mx-1">{{ page }}</span>
             {% else %}
                 <button class="btn btn-outline-secondary mx-1"
-                    hx-get="{% url 'render_table' %}?page={{ page }}&{{ url_encoded_search }}&search_column={{ search_column }}"
+                    hx-get="{% url 'render_table' %}?page={{ page }}"
+                    hx-include="#table-filters-form"
                     hx-target="#table-container">
                     {{ page }}
                 </button>
@@ -45,7 +48,8 @@
         {% endfor %}
         <button class="btn btn-primary"
             {% if page_obj.has_next %}
-                hx-get="{% url 'render_table' %}?page={{ page_obj.next_page_number }}&{{ url_encoded_search }}&search_column={{ search_column }}"
+                hx-get="{% url 'render_table' %}?page={{ page_obj.next_page_number }}"
+                hx-include="#table-filters-form"
                 hx-target="#table-container"
             {% else %} disabled {% endif %}>
             Next
--- a/ftva_lab_data/templates/partials/pagination_controls.html
+++ b/ftva_lab_data/templates/partials/pagination_controls.html
@@ -10,7 +10,7 @@
                 name="items_per_page"
                 class="form-select form-select-md"
                 style="width: fit-content;"
-                hx-get="{% url 'render_table' %}?search={{ search }}&search_column={{ search_column }}"
+                hx-get="{% url 'render_table' %}?{{ url_encoded_search }}&search_column={{ search_column }}"
                 hx-target="#table-container"    
             >
                 {% for option in items_per_page_options %}
@@ -25,7 +25,7 @@
     <div class="col d-flex justify-content-center align-items-center gap-2">
         <button class="btn btn-primary"
             {% if page_obj.has_previous %}
-                hx-get="{% url 'render_table' %}?page={{ page_obj.previous_page_number }}&search={{ search }}&search_column={{ search_column }}"
+                hx-get="{% url 'render_table' %}?page={{ page_obj.previous_page_number }}&{{ url_encoded_search }}&search_column={{ search_column }}"
                 hx-target="#table-container"
             {% else %} disabled {% endif %}>
             Previous
@@ -37,7 +37,7 @@
                 <span class="btn btn-secondary active mx-1">{{ page }}</span>
             {% else %}
                 <button class="btn btn-outline-secondary mx-1"
-                    hx-get="{% url 'render_table' %}?page={{ page }}&search={{ search }}&search_column={{ search_column }}"
+                    hx-get="{% url 'render_table' %}?page={{ page }}&{{ url_encoded_search }}&search_column={{ search_column }}"
                     hx-target="#table-container">
                     {{ page }}
                 </button>
@@ -45,7 +45,7 @@
         {% endfor %}
         <button class="btn btn-primary"
             {% if page_obj.has_next %}
-                hx-get="{% url 'render_table' %}?page={{ page_obj.next_page_number }}&search={{ search }}&search_column={{ search_column }}"
+                hx-get="{% url 'render_table' %}?page={{ page_obj.next_page_number }}&{{ url_encoded_search }}&search_column={{ search_column }}"
                 hx-target="#table-container"
             {% else %} disabled {% endif %}>
             Next
--- a/ftva_lab_data/views.py
+++ b/ftva_lab_data/views.py
@@ -21,6 +21,7 @@ from .views_utils import (
     get_search_result_items,
     get_items_per_page_options,
     format_data_for_export,
+    build_url_parameters,
 )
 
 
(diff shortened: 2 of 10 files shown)


ftva_lab_data/templates/partials/pagination_controls.html
```diff
@@ -10,7 +10,7 @@
                 name="items_per_page"
                 class="form-select form-select-md"
                 style="width: fit-content;"
-                hx-get="{% url 'render_table' %}?search={{ search }}&search_column={{ search_column }}"
+                hx-get="{% url 'render_table' %}?{{ url_encoded_search }}&search_column={{ search_column }}"
```
akohler, line 13: I think it's time to simplify these (this one, and all the others) and their references to multiple parameters.

Let's have the view contexts pass one variable containing (encoded) search, search_column, page (when relevant), and any other URL parameters required.  Call it `url_parameters` or something similar.  Each view knows what needs to be passed to its template, and the templates don't need to perform that verbose concatenation in multiple places.  From what I see, the parameters in each context/template are used consistently - we don't have cases where a single template is building a URL with `page` sometimes and without it others.

This is what I used to check, and I think the data supports my suggestion:
```
~/projects/ftva-lab-data/ftva_lab_data/templates$ grep url_encoded_search -r
search_results.html:     hx-get="{% url 'render_table' %}?{{ url_encoded_search }}&search_column={{ search_column }}&page={{ page }}"
add_edit_item.html:            href="{% if item %}{% url 'view_item' item.id %}{% else %}{% url 'search_results' %}{% endif %}?{{ url_encoded_search }}&search_column={{ search_column }}&page={{ page }}">
view_item.html:            href="{% url 'edit_item' header_info.id %}?{{ url_encoded_search }}&search_column={{ search_column }}&page={{ page }}">
view_item.html:        href="{% url 'search_results' %}?{{ url_encoded_search }}&search_column={{ search_column }}&page={{ page }}">
partials/search_results_table.html:                            href="{% url 'view_item' row.id %}?{{ url_encoded_search }}&search_column={{ search_column }}&page={{ page_obj.number }}">
partials/pagination_controls.html:                hx-get="{% url 'render_table' %}?{{ url_encoded_search }}&search_column={{ search_column }}"
partials/pagination_controls.html:                hx-get="{% url 'render_table' %}?page={{ page_obj.previous_page_number }}&{{ url_encoded_search }}&search_column={{ search_column }}"
partials/pagination_controls.html:                    hx-get="{% url 'render_table' %}?page={{ page }}&{{ url_encoded_search }}&search_column={{ search_column }}"
partials/pagination_controls.html:                hx-get="{% url 'render_table' %}?page={{ page_obj.next_page_number }}&{{ url_encoded_search }}&search_column={{ search_column }}"
```
`page` might be an exception, as I see a couple of these where `page={{ page_obj.number }}` needs to be determined within the template.  If the URL parameters vary within a template, this could still work by building `url_parameters` for what is consistent, and appending only the ones that differ within the template.

ftva_lab_data/views.py
```diff
@@ -90,6 +91,8 @@ def edit_item(request: HttpRequest, item_id: int) -> HttpResponse:
         "title": "Edit Item",
         "button_text": "Save Changes",
         "search": search,
+        # Encode search string here to make it safe for use in URL query strings
+        "url_encoded_search": urlencode({"search": search}),
```
akohler, line 95: See my comment about simplifying the templates by moving parameter concatenation to the views, and passing it via the contexts.

Since there's repeated code here too, with something similar happening in many views, pull it into a simple method in `view_utils.py`.  I'd suggest something like `build_url_parameters(**kwargs)` or similar, which just unpacks `**kwargs`, applies special handling if needed... though it seems like the entire list of URL parameters could be encoded (rather, the concatenated string could be encoded).  

That's a simple way to make future changes easier.  If we need to add another parameter, which may or not require encoding, the logic is all in one place and may require no changes.

akohler (changes requested):
@henry-r18 This does fix the search problems, all known cases and should be general enough to handle similar unknown ones.  

I added a couple of comments requesting some refactoring of the template code, with supporting changes in the views and contexts (and tests, I believe, in `TablePaginationTestCase()`).  I think that changes like these will make the code easier to read and maintain; if you find differently during implementation (code vs PR comments...), we can revisit this.

I think tests should be added, if practical.  They probably could fit into `TablePaginationTestCase()`, which tests UI interaction with search.  It might be as simple as asserting that the URL requested at different points is the same, or that the URL sent to the browser is properly encoded. 

Finally, thank you for the detailed notes about options with HTMX, I appreciate this.  From those, and looking `hx-include`, I agree that deeper integration of some of these things with HTMX would require considerably more work to do right.  At this point, I want to stick with the level of hybrid that we have.  But if the lifespan / scope of this application gets extended significantly... I'll probably be in favor of a 2.0 to go deeper into HTMX and away from Django templates.

If any questions about my requested changes, please let me know.
Thanks --Andy

henry-r18:
@akohler this is ready for another look, when you have a chance. I incorporated your suggest of moving the URL encoding logic to `view_utils.py` using `**kwargs`. That helped to make the views more readable, I think. Now, the views that need the search params encoded receive an encoded query string with the necessary params in their context as `url_parameters`. The exception is the `search_results` view, where I kept the parameters individually defined in the context, as the search string in particular needs to be reflected in the input element.

I'll add some inline comments to clarify things a bit more, but the short of it is that we now: 1) have the desired functionality relating to reserved characters in URL query strings; and 2) improved readability in our templates somewhat, by avoiding the query string concatenation in the templates and pushing it off to the view.

ftva_lab_data/templates/partials/pagination_controls.html
```diff
@@ -10,7 +10,8 @@
                 name="items_per_page"
                 class="form-select form-select-md"
                 style="width: fit-content;"
-                hx-get="{% url 'render_table' %}?search={{ search }}&search_column={{ search_column }}"
+                hx-get="{% url 'render_table' %}"
+                hx-include="#table-filters-form"
                 hx-target="#table-container"
```
henry-r18, line 15: The pagination controls should always be within the same DOM scope as `#table-filters-form`, so we can just use `hx-include` here.

ftva_lab_data/views.py
```diff
@@ -105,7 +107,7 @@ def edit_item(request: HttpRequest, item_id: int) -> HttpResponse:
             messages.success(request, "Item updated successfully!")
             url = (
                 f"{reverse('view_item', args=[item.id])}"
-                f"?search={search}&search_column={search_column}&page={page}"
+                f"?{urlencode({"search": search})}&search_column={search_column}&page={page}"
```
akohler, line 110: I think you can leverage the new `build_url_parameters()` here too.

akohler (changes requested):
@henry-r18 Thank you for the refactoring, and for the added comments, especially around use of `hx-include="#table-filters-form"` which simplifies the templates even further.  I think this helps make our hybrid use of HTMX + Django clearer.

I noted one place, in the `edit_item` view when a record is updated, URL parameters are being encoded within the view instead of via `build_url_parameters()`.  I think this can be changed as well; if not, please document the reason.

Once that's resolved, I'll merge.

Thanks --Andy

henry-r18:
Thanks for catching that, @akohler! Oversight on my part--no other reason.

Just pushed the change to `views.py` (as well as removing the now unused import of `urlencode`). I also went ahead and rebased on top of Zoe's (@ztucker4) recently merged work.

akohler (approved):
@henry-r18 All good now, merging.  Thanks --Andy
@ztucker4 FYI.

PR #56, opened 2025-07-29
SYS-1935: Add mgmt command to extract missing inventory numbers

henry-r18:
Implements [SYS-1935](https://uclalibrary.atlassian.net/browse/SYS-1935)

### Acceptance criteria
- [ ] Updates records in the Digital Data application which do not currently have inventory numbers, but for which a valid inventory number exists in the concatenation of the `file_folder_name`, `sub_folder_name`, and `file_name` fields
- [ ] Should be a Django management command, to fit cleanly into that environment
- [ ] Code should be copied from the original `extract_inventory_number.py` script in the `ftva-mams-data` repository
- [ ] The command must have a `--dry-run` flag and generate accurate detailed preview data about what would be changed, without changing data

### Description
This PR adds a management command to extract inventory numbers from a concatenation of the `file_folder_name`, `sub_folder_name`, and `file_name` fields. The script utilizes code adapted from the `extract_inventory_number.py` script in the `ftva-mams-data` repo. A `--dry_run` flag is included to produce a summary in the form of a CSV file of records that would be updated in live mode, without actually updating anything. The summary is also created in live mode, but without the dry run indicators in the filename and first row of the CSV.

In my local environment, the script identifies 4041 records where a matching inventory number exists in the path-like fields, but where the `inventory_number` field is either empty or contains the value "invalid". The resulting CSV file can be shared with FTVA staff for their reference to confirm these changes should be applied in production.

To run the command in dry run mode within the dev container, run:

`python manage.py extract_inventory_numbers --dry_run`

### Testing
I added no new automated tests, as the regex pattern has been validated elsewhere, and we can trust the Django methods work. Please confirm that the number of records found is reasonable and that the updates are applied correctly in live mode.


henry-r18:
@akohler This is ready for another look, when you have a chance.

I incorporated changes to the regular expression discussed with Thelma on Slack and added a test class and some new cases accordingly. I also refactored `extract_inventory_numbers.py` to better encapsulate the logic for removing false positives and preparing the output string.

If the script looks good to you, would you be able to send the latest production data dump from the DD app? That way, I can give Thelma an up-to-date summary file to review.

henry-r18:
@akohler I tidied up the comments and added a few more test cases intended to address the new specs related to single-letter prefixes. I noticed that the substring `M4V` (or `M4v`) appears several times in the path fields, and I think it's safe to assume this refers to the file format, so I went ahead and added `M4` to the false-positives list. When grepping around, I don't see any `M4` sub-strings that look like they're genuine inventory numbers, though if there are any, they can be treated manually. Let me know if you think of any more cases that need testing.

akohler:
@henry-r18 I ran the updated command against the prod data I just shared.  About 500 fewer records checked this time, which is interesting - presumably some cleanup already done in production, more than regex changes?

Excluding `M4` makes sense, as it did not seem to be a legitimate inventory number from earlier conversation.

I think send the full current CSV of proposed changes to Thelma for review, and if she says OK tomorrow morning, I'll give this a final review and merge.

akohler (approved):
@henry-r18 This is great, thanks for making the changes and working to refine & resolve the data differences.  Merging now, thanks.

@ztucker4 FYI.

PR #57, opened 2025-07-30
SYS-1942: add record retrieval endpoint w/HTTP basic auth

ztucker4:
Implements [SYS-1942](https://uclalibrary.atlassian.net/browse/SYS-1942)

Adds a new endpoint, `/records/<record_id:int>`, that retrieves JSON data for an individual `SheetImport` object.

The endpoint uses HTTP basic auth, and requires valid Django account credentials (username and password for an active user). To support this, I added a new `basic_auth_required` decorator, defined in `views_utils.py`.

Given that this data will be used outside of this application, I wasn't sure whether to include Status and Assigned User data in the output JSON. I opted to include both for now.

Two simple success/failure tests are added in `tests.py`, for a total of 53: `python manage.py test`

[SYS-1942]: https://uclalibrary.atlassian.net/browse/SYS-1942?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/ftva_lab_data/tests.py
+++ b/ftva_lab_data/tests.py
@@ -33,6 +33,7 @@ from ftva_lab_data.views_utils import (
     build_url_parameters,
 )
 from ftva_lab_data.table_config import COLUMNS
+import base64
 
 
 class GetFieldValueTestCase(TestCase):
@@ -196,6 +197,34 @@ class UserAccessTestCase(TestCase):
         # Confirm item successfully assigned to user
         self.test_object.refresh_from_db()
         self.assertEqual(self.test_object.assigned_user, self.authorized_user)
+
+    def test_basic_auth_user_can_get_record(self):
+        """Asserts that a user with basic auth can GET a record."""
+        # Construct HTTP Basic Auth credentials
+        credentials = "authorized:testpassword"
+        base64_credentials = base64.b64encode(credentials.encode("utf-8")).decode(
+            "utf-8"
+        )
+        url = reverse("get_record", args=[self.test_object.id])
+        response = self.client.get(
+            url, HTTP_AUTHORIZATION=f"Basic {base64_credentials}"
+        )
+        self.assertEqual(response.status_code, 200)
+        # Confirm the response contains the expected data
+        self.assertIn(self.test_object.file_name.encode(), response.content)
+
+    def test_basic_auth_bad_credentials(self):
+        """Asserts that a user with bad basic auth credentials receives 401."""
+        # Construct HTTP Basic Auth credentials with incorrect password
+        credentials = "authorized:wrongpassword"
+        base64_credentials = base64.b64encode(credentials.encode("utf-8")).decode(
+            "utf-8"
+        )
+        url = reverse("get_record", args=[self.test_object.id])
+        response = self.client.get(
+            url, HTTP_AUTHORIZATION=f"Basic {base64_credentials}"
+        )
+        self.assertEqual(response.status_code, 401)
 
 
 class TablePaginationTestCase(TestCase):
--- a/ftva_lab_data/tests.py
+++ b/ftva_lab_data/tests.py
@@ -202,9 +202,7 @@ class UserAccessTestCase(TestCase):
         """Asserts that a user with basic auth can GET a record."""
         # Construct HTTP Basic Auth credentials
         credentials = "authorized:testpassword"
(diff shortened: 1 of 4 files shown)


ftva_lab_data/views_utils.py
```diff
@@ -288,3 +291,37 @@ def build_url_parameters(**kwargs) -> str:
...
+                return HttpResponse("Invalid basic auth header", status=400)
+            # Authenticate using Django's authenticate function
+            user = authenticate(request, username=username, password=password)
+            if user is not None and user.is_active:
+                request.user = user
+                login(request, user)
+                return view_function(request, *args, **kwargs)
+        # If not authenticated, prompt for login
+        response = HttpResponse("Unauthorized", status=401)
+        # Set the WWW-Authenticate header to prompt for basic auth
```
akohler, line 323: I'm confused by this.  Given the flow, my first reading was that this would always add the `WWW-Authenticate` header.  Looking more carefully, I see that successful authentication (via the request header) returns before this.

So it does work, but the method is long enough that this would benefit from an `else`.  If the expected header-based authentication didn't work, then do the unauthorized response prompt and return that from inside the `else`.

ftva_lab_data/views.py
```diff
@@ -365,3 +366,30 @@ def release_notes(request: HttpRequest) -> HttpResponse:
     :return: Rendered HTML for the release notes.
     """
     return render(request, "release_notes.html")
+
+
+@basic_auth_required
+def get_record(request: HttpRequest, record_id: int) -> HttpResponse:
```
akohler, line 372: Returns `JsonResponse` :)

ftva_lab_data/tests.py
```diff
@@ -197,6 +198,34 @@ def authorized_user_can_assign_user(self):
...
+        base64_credentials = base64.b64encode(credentials.encode("utf-8")).decode(
+            "utf-8"
+        )
+        url = reverse("get_record", args=[self.test_object.id])
+        response = self.client.get(
+            url, HTTP_AUTHORIZATION=f"Basic {base64_credentials}"
+        )
+        self.assertEqual(response.status_code, 200)
+        # Confirm the response contains the expected data
+        self.assertIn(self.test_object.file_name.encode(), response.content)
```
akohler, line 214: The call to `.encode()` here is inconsistent with the other 2 calls (lines 205 and 220).  `str.encode()`  has an `encoding` parameter which defaults to `utf-8`, so it's fine to leave it out, or specify it in all calls.  Same default with the `.decode()`.

I vote to leave it out, as the default.

akohler (changes requested):
@ztucker4 This is excellent! Well done for providing a custom decorator.  I assumed Django would handle that already, but (checks documentation) I see it does not.  Re-looking at the Django REST framework, I still think it's overkill for this one view, but definitely something to keep in mind if other projects need to expose more endpoints.

I flagged 2 small changes in the decorator, for more specific exception handling and more explicit flow control.  A couple of other very small tweaks, and a non-review comment about type hints for callables in case you want to read more, but no change needed there.

Tests are appropriate and all pass.  I verified functionality by visiting http://127.0.0.1:8000/records/12345/ - the first request, even though signed in (via Django itself) already, I was prompted to go through basic auth anyhow, which I did.  Second request, I was not prompted.  Since our use case is almost certainly just programmatic API calls, I don't think this matters.

I also tested via `curl`, both unauthenticated and with default dev admin credentials, and got the correct responses in both cases.

Thanks --Andy'

ztucker4:
Hi @akohler , this is ready for another look. All comments (except type hints for the decorator) are addressed. Moving the 401 response in the decorator to an `else` block required creating a second `else` block to handle the case of correctly formed Basic Auth headers that still fail to authenticate.

akohler (approved):
@ztucker4 Thanks for the updates.  I didn't like the repeated 401-related code in the decorator, so I refactored a bit.  All tests still pass, automated and manual, but please take a look and if the change is OK with you, go ahead and merge.

Thanks --Andy

ztucker4:
Your change looks good to me, thanks! Merging now.

akohler:
@ztucker4 Thanks.
@henry-r18 FYI, this merge affects `tests.py`.

PR #60, opened 2025-08-01
SYS-1945: Retrieve FM records for an inventory number

henry-r18:
Implements [SYS-1945](https://uclalibrary.atlassian.net/browse/SYS-1945)

### Acceptance criteria
- [x] Write a view which accepts an inventory_number (which will come from the specific DD record the user is on) and uses FilemakerClientto retrieve matching records from Filemaker.

    - [ ] Limit the returned records to 10: either take the first 10 retrieved, or we probably change the client (via separate ticket) to accept a limit.

- [x] For each Filemaker record found, get the specific fields needed (see [list from Thelma](https://uclalibrary.atlassian.net/wiki/x/EwDMTQ))

- [x] Create a search results list of dictionaries, which can be passed to a generic template (to be built later) for display.

    - [ ] If needed, restructure the field data  as key-value pairs which can be passed to a generic template (to be built later) for display.

### Description
This PR adds a new view function: `get_filemaker_data()`. The function accepts an inventory number and returns a list of objects which include Filemaker records matching that inventory number. Only the specific fields that FTVA defined are returned in the `full_data` field of the search result object.

The function mimics the `get_alma_data()` view @ztucker4 added earlier. One difference is that FM credentials are loaded from environment variables using `os.getenv()` in `settings.py`. These are used to establish the FM client instance.

I wasn't sure if I should add the 10-record limit here, or if we'd prefer to do that in `ftva-etl`, so I left that off for now.

I opted not to "prettify" the keys returned in the `full_data` dict quite yet, as I'm not sure what FTVA would prefer, as there's already a mixed style in the field names returned from Filemaker.

### Environment variables
The following environment variables will need to be defined for authentication with the FM client:
```
FILEMAKER_USER={USERNAME HERE}
FILEMAKER_PASSWORD={PW HERE}
```

### Testing
I haven't yet added tests, but can do, if there are certain cases we want to test against.

[SYS-1945]: https://uclalibrary.atlassian.net/browse/SYS-1945?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/ftva_lab_data/views.py
+++ b/ftva_lab_data/views.py
@@ -7,7 +7,7 @@ from django.http import HttpRequest, HttpResponse, StreamingHttpResponse, JsonRe
 from django.shortcuts import render, redirect
 from django.template.loader import render_to_string
 from django.urls import reverse
-from ftva_etl import AlmaSRUClient
+from ftva_etl import AlmaSRUClient, FilemakerClient
 import pandas as pd
 import io
 
@@ -24,6 +24,7 @@ from .views_utils import (
     build_url_parameters,
     basic_auth_required,
     process_full_alma_data,
+    get_specific_filemaker_fields,
 )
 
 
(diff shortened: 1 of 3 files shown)


ftva_lab_data/views.py
```diff
@@ -432,3 +433,57 @@ def get_alma_data(request: HttpRequest, inventory_number: str) -> list[dict]:
...
+        "element_info",
+        "spac",
+        "episode no.",
+        "film base",
+        "donor_code",
+    ]
+    full_data_dicts = []
+
+    for record in records:
+        record = get_specific_filemaker_fields(record, filemaker_fields)
```
akohler, line 478: I'd like it to be clearer that this is a set of filtered fields, especially as the code in the next few lines is acting on this reduced record, not the original one in the `for record in records` loop.  The fact that `record` changes from a (Filemker) `Record` on line 477 to a `dict` on line 478 adds potential for confusion.  Since the method is `get_specific_filemaker_fields`, rename the variable to `filemaker_fields` or `fm_fields` to help make this clearer.

ftva_lab_data/views_utils.py
```diff
@@ -396,3 +397,20 @@ def process_full_alma_data(field_list: list[Field]) -> dict[str, str]:
...
+
+    :param Record fm_record: A fmrest Record instance.
+    :param list[str] specific_fields: A list of specific fields to get from the Record.
+    :return: A dict with the specific fields from the Filemaker Record.
+    Fields are only included if they exist in the Record.
+    """
+    return {
+        field: fm_record[field]
+        for field in specific_fields
+        if field in fm_record.to_dict()
```
akohler, line 415: This is nice and Pythonic, which I like :) - but I believe the call to `fm_record.to_dict()` occurs with every repetition in the comprehension.  That could be an expensive repeated call (though here's it's probably not too bad).  But it can be avoided, and the code made a little clearer, by doing:
```
record_fields = fm_record.to_dict()
return {.... if field in record_fields}
```

My little test in the REPL to try to show this:
```
>>> def foo():
...   print("in foo")
...   return {"a": 1, "b": 2}
...
>>> foo()
in foo
{'a': 1, 'b': 2}

>>> reps = ["a", "b", "c"]

>>> ["x" for val in reps if val in foo()]
in foo
in foo
in foo
```
The message shows the function is called for each value in the tested `reps`.

I don't know if Python optimizes this loop - since there are no parameters, it could easily do so - but I think this change makes the code's intent clearer and avoids possible performance problems.

akohler (changes requested):
@henry-r18 This works as expected: I tested with inventory numbers to return 0, 1, 2, and 100 records.  All worked without problems and showed me the relevant data.

Processing all returned records is fine for now.  The package client is where this should go, as a parameter the caller can set / override as needed.  Also no need for tests on this code as no real data transformation is occurring.

I do want the raw FM field names transformed to be consistent, both with themselves and with the basic Alma "Field 245" etc.  Users of the FM real client interface don't see the raw names, just labels assigned by whoever created the layouts, and even this handful of fields shows the, um, creativity of the FM database.  
For now, a basic transformation is fine:
* Replace underscores with spaces
* Use sentence case
  * One exception, in these fields: change "spac" to "SPAC" (an UCLA library acronym, for [SPecial collections and Archives Codes](https://uclalibrary.atlassian.net/wiki/x/6oGUBQ))

Other than that, 2 requested minor code changes for clarity, then this will be ready to merge.

Thanks --Andy

PR #61, opened 2025-08-01
SYS-1946: Display external search results

ztucker4:
Implements [SYS-1946](https://uclalibrary.atlassian.net/browse/SYS-1946)

Adds a new template, `external_search_results.html`, with a corresponding view. The template is intended to support both Alma and Filemaker data.

Items with an Inventory Number value should have a two new buttons visible in the top right for Alma and Filemaker search results. Both buttons should open a new tab when clicked, and the Alma button should display search results. 

image
image

Note for the FM implementation: The FM view will need to return an `inventory_number` and a `search_type` in its context, as these are used in the template and the generic `get_external_search_results` view. The view assumes the `search_type` is either "alma" or "fm".

### DIFF
--- a/ftva_lab_data/templates/external_search_results.html
+++ b/ftva_lab_data/templates/external_search_results.html
@@ -1,0 +1,43 @@
+{% extends 'base.html' %}
+
+{% block content %}
+
+<div class="w-75 mx-auto">
+    {% if search_type == "alma" %}
+    <h1>Alma Search Results</h1>
+    {% else %}
+    <h1>Filemaker Search Results</h1>
+    {% endif %}
+</div>
+
+<div class ="d-flex align-items-stretch gap-2 mb-3 w-75 mx-auto">
+    <div class="w-100">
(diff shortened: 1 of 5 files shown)


akohler (commented):
@ztucker4 This looks great!  I tried with several records, and it immediately makes it (a) easier to see what's in Alma for a DD record and (b) how many DD records with inventory numbers don't seem to have matches in Alma :( - but that's out of our scope.

I don't have any significant concerns with this code, though I expect we'll make minor tweaks as we integrate everything.  The only reason I'm not merging now is I think it will be cleaner to merge @henry-r18 's work on the FM retrieval first.  Then I'll modify this to incorporate his changes.

Thanks --Andy

akohler (approved):
Meant to make my review an approval, doing it now.
I will  merge this after #60 is merged.

akohler:
@henry-r18 FYI, search results page has been merged and includes FM search results as well as Alma, for rebase into your WIP.

PR #62, opened 2025-08-06
SYS-1947: Add record details modal to external search results view

henry-r18:
Implements [SYS-1947](https://uclalibrary.atlassian.net/browse/SYS-1947)

### Acceptance criteria
- [x] Create a template which displays a record from Alma or Filemaker (see linked tickets).
- [x] The template should display a simple list of the key / value pairs from the dictionary.
- [x] The template will display as an overlay on the search results page (see linked ticket), as a bootstrap scrollable overlay (I think we decided).
- [x] Closing the template will leave the user viewing the same search results page as when the template was opened.  No other navigation is required.

### Description
This PR adds a record detail modal to the external search results display. The user can now click on the `View Record` button(s) within the external search results table and launch a modal with the key-value pairs for that record displayed in a scrollable overlay, provided by Bootstrap.

Markup for all the modals is loaded with the initial page load. This should be acceptable for the number of records we're dealing with, keeping all data provided by a single view function and reducing the need to involve HTMX. It does however push some logic into the template. I think it is an acceptable amount of logic in the template, mostly just involving for loops. The modal markup is loaded using Django's `{% include %}`, passing two variables in that statement: the record for the current table row and a modal index that allows Bootstrap to launch the right modal for the record at hand.

Let me know if this is too much logic to put into the template, and I can find a different approach using a new view and HTMX.

[SYS-1947]: https://uclalibrary.atlassian.net/browse/SYS-1947?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/ftva_lab_data/templates/external_search_results.html
+++ b/ftva_lab_data/templates/external_search_results.html
@@ -30,9 +30,20 @@
                 <tr>
                     <td>{{ record.record_id }}</td>
                     <td>{{ record.title }}</td>
-                    <!--TODO: button should display external item details-->
-                    <td><a href=# class="btn btn-sm btn-primary">View Record</a></td> 
+                    <td>
+                        <button 
+                            type="button" 
+                            class="btn btn-sm btn-primary"
+                            data-bs-toggle="modal"
+                            data-bs-target="#record-modal-{{ forloop.counter }}">
+                            View Record
+                        </button>
+                    </td> 
                 </tr>
+                {% comment %} Include a modal for each record, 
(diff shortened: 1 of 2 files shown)


akohler (approved):
@henry-r18 This is great!  The display is clean, the template code is simple.  I think it's perfect for our initial release. 

Merging now, thanks --Andy

PR #63, opened 2025-08-07
SYS-1949: Add view to generate metadata JSON

henry-r18:
Implements [SYS-1949](https://uclalibrary.atlassian.net/browse/SYS-1949)

### Acceptance criteria
- [x] A new view takes the current record’s inventory_numberas a parameter.
- [x] The view uses the `ftva_etl` package to retrieve both Alma and Filemaker records with that inventory number.
- [x] If both Alma and Filemaker searches return exactly one record each, pass Alma / Filemaker / Digital Data records to the ETL package’s `get_mams_metadata()` function.
- [x] If either Alma or Filemaker searches returns other than exactly one record each, return a dictionary with useful error message(s), for now, maybe like the below.

### Description
This PR adds a view to generate a combined metadata record in JSON, using records retrieved from Alma, Filemaker, and Django. The view takes in a `record_id` referring to the Django record to use as the "base" for the metadata generation, then an `inventory_number` to use to retrieve records from Alma and Filemaker. If there is exactly one record each retrieved from Alma and Filemaker, these are joined with the Django record using the `get_mams_metadata()` from the `ftva_etl` package. Otherwise, a message is returned.

`TODO` notes are included for the return statements that will eventually render templates.

### Testing
I didn't add any automated tests, but using the recent production data dump in a local instance, the following URL should provide an example of a 1-1-1 record that returns a joined JSON record: http://localhost:8000/metadata_json/106/M190816/. The Django `record_id` and `inventory_number` are the second-to-last and last URL components respectively.

[SYS-1949]: https://uclalibrary.atlassian.net/browse/SYS-1949?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/ftva_lab_data/urls.py
+++ b/ftva_lab_data/urls.py
@@ -22,4 +22,9 @@ urlpatterns = [
         views.get_external_search_results,
         name="get_external_search_results",
     ),
+    path(
+        "metadata_json/<int:record_id>/<str:inventory_number>/",
+        views.generate_metadata_json,
+        name="generate_metadata_json",
+    ),
 ]
--- a/ftva_lab_data/urls.py
+++ b/ftva_lab_data/urls.py
@@ -23,7 +23,7 @@ urlpatterns = [
         name="get_external_search_results",
     ),
     path(
-        "metadata_json/<int:record_id>/<str:inventory_number>/",
+        "metadata_json/<int:record_id>/",
         views.generate_metadata_json,
         name="generate_metadata_json",
     ),
--- a/ftva_lab_data/views.py
+++ b/ftva_lab_data/views.py
@@ -7,7 +7,7 @@ from django.http import HttpRequest, HttpResponse, StreamingHttpResponse, JsonRe
 from django.shortcuts import render, redirect
 from django.template.loader import render_to_string
 from django.urls import reverse
-from ftva_etl import AlmaSRUClient, FilemakerClient
+from ftva_etl import AlmaSRUClient, FilemakerClient, get_mams_metadata
 import pandas as pd
 import io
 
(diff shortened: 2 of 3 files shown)


ftva_lab_data/urls.py
```diff
@@ -22,4 +22,9 @@
         views.get_external_search_results,
         name="get_external_search_results",
     ),
+    path(
+        "metadata_json/<int:record_id>/<str:inventory_number>/",
```
akohler, line 26: I think the URL should not include the inventory number. The view already has to retrieve the record itself, and can easily pull the inventory number out of that.  Removing it from the URL also simplifies what the calling template has to do.

ftva_lab_data/views.py
```diff
@@ -524,3 +524,51 @@ def get_external_search_results(
...
+    """Generate a  JSON metadata record for a given inventory number,
+    by combining data from Alma, Filemaker, and Django.
+
+    :param request: The HTTP request object.
+    :param record_id: The ID of the Django record to use.
+    :param inventory_number: The inventory number to search for in Alma and Filemaker.
+    :return: A JSON record containing the combined metadata.
+    """
+
+    # Get Django record
```
akohler, line 541: The logic to get the record here (the dict required by the metadata function) should be the same as the logic to get it via the REST-ish API call used by outside callers.
This logic currently is most of the code in `views.get_record()`, so time to pull that into a separate method in `views_utils`.  That way, Django data passed to the metadata function for a given record will be identical, regardless of caller.

ftva_lab_data/views.py
```diff
@@ -524,3 +524,51 @@ def get_external_search_results(
...
+        return JsonResponse(metadata)
+
+    # Otherwise, Alma and/or FM records are either not unique or not found,
+    # so return a message with the counts of records found
+    message = (
+        f"Metadata not generated because the search for inventory number {inventory_number} "
+        f"found {bib_records_count} Alma and {fm_records_count} Filemaker records."
+    )
+    # TODO: render a template with the message. Returning JSON for now.
+    return JsonResponse({"message": message}, status=400)
```
akohler, line 574: Since this view is not intended to be called from outside of Django, the `status` can be removed - which it probably would be when switching to a template anyhow.

ftva_lab_data/views.py
```diff
@@ -524,3 +513,50 @@ def get_external_search_results(
...
+def generate_metadata_json(request: HttpRequest, record_id: int) -> JsonResponse | str:
+    """Generate a  JSON metadata record for a given inventory number,
+    by combining data from Alma, Filemaker, and Django.
+
+    :param request: The HTTP request object.
+    :param record_id: The ID of the Django record to use.
+    :return: A JSON record containing the combined metadata.
+    """
+
+    django_record_data = transform_record_to_dict(record_id)
```
henry-r18, line 527: Do you think we need to handle potential `SheetImport.DoesNotExist` errors here? This view should really only be called from the `view_item` template, which should ensure the record exists, but I suppose it would be safest to handle it like the `get_record` view does.

akohler: Probably the try/except is not needed, as you say the context is an existing record being viewed, but no harm so let's keep it.

PR #66, opened 2025-08-14
SYS-1963: add file_type field to SheetImport objects

ztucker4:
Implements [SYS-1963](https://uclalibrary.atlassian.net/browse/SYS-1963)

Adds a new `file_type` field to `SheetImport` objects. This field is constrained to the list of values Christina provided on the Jira ticket. This is presented as a dropdown list on Add/Edit Item pages, and displayed under File Name on View Item pages.

I assumed that these values won't be changing very often, so I chose to hard-code them within `models.py`. 

I also saw that you mentioned users typing values into this field. The current implementation (using a plain `<select>` element) provides basic support for this, but let me know if you think a more complete typeahead feature is needed.

image
image

[SYS-1963]: https://uclalibrary.atlassian.net/browse/SYS-1963?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


akohler (approved):
@ztucker4 Your implementation is fine with me.  Putting the choices on the model does require a migration if future changes are needed, but that's trivial enough, and for this application I think enforcing choices at the database level is appropriate...
No type-ahead needed, this specific order and UX is what was requested.  We can make changes later if they change their minds.

Merging now, thanks --Andy
@henry-r18 FYI

PR #67, opened 2025-08-14
SYS-1962: Add UUIDField to SheetImport model

henry-r18:
Implements [SYS-1962](https://uclalibrary.atlassian.net/browse/SYS-1962)

### Acceptance criteria
- [x] Add UUID field to `SheetImport` model, using `uuid.uuid4()`
- [x] New records get UUID populated on creation
- [x] Existing records get UUID populated via migration
- [x] UUID displays below Inventory Number in the Inventory Information area of the view page
- [x] UUID displays next to record id in the top area when editing
- [x] UUID cannot be edited

### Description
This PR adds UUIDs to the `SheetImport` model and to all existing records via the included DB migration.

New `SheetImport` records get a UUID using the `uuid4()` module by default, and the field is set to `editable=False` on the model. The view- and edit-item pages now display the UUID as described in the acceptance criteria.

The migration includes a custom function to generate UUIDs for all existing records that don't have them. This will be applied when the migration is run via `python manage.py migrate` from within the dev container. One potential issue is that we'll all get different UUIDs in our local systems, until we can synchronize with the production data. I suppose that can be done by dumping the `SheetImport` records, like we've done in the past.

### Testing
I can add a test that demonstrates new `SheetImport` records get a UUIDv4 by default, if you'd like. This can also be manually tested by creating a new test record via the UI.

**Update**: I had initially forgotten to run tests, and recent changes caused tests to fail, as the `_get_combined_field_data` function of the `clean_imported_data.py` management command was attempting to concatenate the new UUID field with strings. I added the UUID to the list of fields excluded from this function, and now all 57 existing tests are passing again.

image

image

[SYS-1962]: https://uclalibrary.atlassian.net/browse/SYS-1962?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/charts/prod-ftvalabdata-values.yaml
+++ b/charts/prod-ftvalabdata-values.yaml
@@ -6,7 +6,7 @@ replicaCount: 1
 
 image:
   repository: uclalibrary/ftva-lab-data
-  tag: v1.0.7
+  tag: v1.0.8
   pullPolicy: Always
 
 nameOverride: ""
--- a/ftva_lab_data/templates/release_notes.html
+++ b/ftva_lab_data/templates/release_notes.html
@@ -4,6 +4,13 @@
 <h3>Release Notes</h3>
 <hr/>
 
+<h4>1.0.8</h4>
+<p><i>August 14, 2025</i></p>
+<ul>
+    <li>Add File Type field with a dropdown of valid values.</li>
+    <li>Add UUID field that is auto-generated for new records.</li>
+</ul>
+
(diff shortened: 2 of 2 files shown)


henry-r18:
@akohler Just bumped the chart version and updated the release notes. Thanks for the reminder!

PR #68, opened 2025-08-20
SYS-1972: Add field for Asset Type

akohler:
Implements [SYS-1972](https://uclalibrary.atlassian.net/browse/SYS-1972). Not tagged for release, will be bundled with other updates.

This PR adds an "Asset Type" field, exposed via UI as a drop-down list with values in the specified order.  As with "File Type", this is implemented as a list of choices directly in the `SheetImport` model.

The field appears directly below "File Type" for viewing and editing.

No new tests; all existing tests pass.
dd_asset_type_edit
dd_asset_type_list
dd_asset_type_view


ztucker4 (approved):
Looks good to me! The new field contains the correct values and is displayed correctly on View and Add/Edit displays. Thanks for updating `convert_dl_sheet_data` with the last few new fields as well.

PR #69, opened 2025-08-20
SYS-1973: Add No Ingest Reason field

akohler:
Implements [SYS-1973](https://uclalibrary.atlassian.net/browse/SYS-1973).  Bumps version to `v1.0.9` for release.

This PR adds an "No Ingest Reason" field, exposed via UI as a drop-down list with values in the specified order. As with "File Type", this is implemented as a list of choices directly in the `SheetImport` model.

The field appears directly below "Date of Ingest" for viewing and editing.

No new tests; all existing tests pass.

dd_no_ingest_reason_edit

dd_no_ingest_reason_list

dd_no_ingest_reason_view


ztucker4 (approved):
Looks good! The field shows up in the right places with the right set of values, and everything looks correct for deployment. Merging now!

PR #70, opened 2025-08-21
SYS-1977: Fix metadata display bug caused by UUID

akohler:
Fixes [SYS-1977](https://uclalibrary.atlassian.net/browse/SYS-1977).  Tagged as `v1.0.10` for deployment.

This PR fixes a bug which prevented the JSON metadata from displaying.  The bug was caused by the `UUID()` data type being included in the Django dictionary generated by `views_utils.transform_record_to_dict()`.  The `uuid` field, which has a type of `UUID()`, could not be serialized to JSON by `json.dumps()`, raising a `TypeError`.

Since we only need the value and not the object itself, I cast `uuid` to a string, which fixes the bug.  I also added a test case to cover this.

`docker compose exec django python manage.py test` now runs 58 tests, all passing.

[SYS-1977]: https://uclalibrary.atlassian.net/browse/SYS-1977?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/charts/prod-ftvalabdata-values.yaml
+++ b/charts/prod-ftvalabdata-values.yaml
@@ -6,7 +6,7 @@ replicaCount: 1
 
 image:
   repository: uclalibrary/ftva-lab-data
-  tag: v1.0.9
+  tag: v1.0.10
   pullPolicy: Always
 
 nameOverride: ""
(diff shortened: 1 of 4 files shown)


henry-r18 (approved):
Thanks @akohler! Confirming that `View Metadata` now functions as expected and all 58 tests pass. Merging now.

image

PR #75, opened 2025-08-29
SYS-1989: UI to batch update carrier location

ztucker4:
Implements [SYS-1989](https://uclalibrary.atlassian.net/browse/SYS-1989)

Adds a new user interface for updating carrier locations in bulk.  Users are shown a simple form, from which they can select an existing Carrier value (taken from all `carrier_a` and `carrier_b` values on `SheetImport` objects in the database) and input any string as a Location value. On submission, the form updates the corresponding `_location` value for all items with that Carrier value in one of the two columns. Users are redirected to the main search results screen and shown a success message with the number of items updated.

The new interface is accessible via the navbar, and requires the same `change_sheetimport` permission needed for editing objects. This implementation uses a new HTMX partial, `carrier_suggestions.html`, with a corresponding view. Javascript is used to allow the form to access selected values, and to enforce that the user selects a valid value from the list before submitting the form.

To test manually, go to the new [Set Carrier Location](http://127.0.0.1:8000/set_carrier_location/) page. Begin typing a carrier value (matches use `icontains`, so case and position in string shouldn't matter), and select one from the list. Type in a Location value and click Set Location. See your updated location in search results.

This PR also includes four new test cases in `tests.py`, for a total of ~~62~~ **66**.

Release notes and deployment tag bump not included - let me know if I should add these.

Screenshots:
image
image
image

[SYS-1989]: https://uclalibrary.atlassian.net/browse/SYS-1989?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


akohler:
@ztucker4 Thank you for the UX changes.  I confirm that this now works as expected without requiring an explicit mouse click to select a carrier: I can navigate the current list with up/down arrows and select with Enter, and this feels much user-friendlier.

The only functional concern I have may be for an edge case: removing locations, which is not in the ticket specs, but occurred to me while testing.  The UI does not allow an empty location field, so I tried entering two quotes `""`... which was accepted, and saved, but as the explicit string `'""'` ... an unexpected result.  I've asked Christina for input.  No changes needed yet, we'll see what she says.

I do wonder about using javascript to remove & add HTMX attributes.  @henry-r18, I've added you as a reviewer.  Can you please take a look at this approach?  Can you suggest an approach which might be simpler?

Once Henry and Christina provide input, I'll review any final changes.  Thanks --Andy

akohler:
@ztucker4 Per discussion on Slack, we should support removing a value, by setting it to the empty string and not, as it currently does, a non-empty string containing `""` or `''` 🤯 
Thanks --Andy

static/js/main.js
```diff
@@ -143,3 +143,107 @@ document.addEventListener("DOMContentLoaded", function () {
...
+    carrierInput.setAttribute("hx-trigger", "keyup changed delay:300ms");
+    carrierInput.setAttribute("hx-target", "#carrier-suggestions");
+    carrierInput.setAttribute("hx-params", "*");
+  }
+
+  // Save the suggestions URL for re-enabling
+  carrierInput.dataset.suggestionsUrl = carrierInput.getAttribute("hx-get");
+
+  // Reset flag and highlight when user types, and re-enable HTMX input
+  carrierInput.addEventListener("input", function () {
```
henry-r18, line 184: @ztucker4 Since this listener is re-setting the value of `carrierSelected` to `false` any time the user changes the `input` field, I think that the `disableCarrierInputTrigger()` and `enableCarrierInputTrigger()` can be consolidated into a single listener for the `htmx:beforeRequest` event that checks if `carrierSelected` is `true`, and if it is, interrupts the request with `e.preventDefault()`. So basically, if a carrier is already selected, HTMX won't create a request.

After removing calls to `disableCarrierInputTrigger()` and `enableCarrierInputTrigger()`, I replaced lines 163 to 181 with the following in my local instance, and it seems to work in the UI, though I could be missing something:

```
 // Prevent HTMX from re-issuing a request after selection
  carrierInput.addEventListener("htmx:beforeRequest", function (e) {
    if (carrierSelected) {
      e.preventDefault();
    }
  });
```

cc// @akohler

ztucker4:
@akohler, this is ready for yet another look. I've added the ability to clear location data from carriers by leaving the Location field blank in the form. Users will see a browser `confirm()` popup if they do this, as demonstrated in the Systems meeting today.

@henry-r18, thanks for recommending listening for the `htmx:beforeRequest` event. Using `preventDefault()` when a carrier is selected works perfectly and is definitely cleaner than my earlier version.

akohler (changes requested):
One problem: checking the history shows no audit information is being captured as locations are updated this way.  I think these changes are important to know about.

The [documentation for update()](https://docs.djangoproject.com/en/5.2/ref/models/querysets/#django.db.models.query.QuerySet.update) confirms:

> Finally, realize that update() does an update at the SQL level and, thus, does not call any save() methods on your models, nor does it emit the [pre_save](https://docs.djangoproject.com/en/5.2/ref/signals/#django.db.models.signals.pre_save) or [post_save](https://docs.djangoproject.com/en/5.2/ref/signals/#django.db.models.signals.post_save) signals (which are a consequence of calling [Model.save()](https://docs.djangoproject.com/en/5.2/ref/models/instances/#django.db.models.Model.save)). If you want to update a bunch of records for a model that has a custom [save()](https://docs.djangoproject.com/en/5.2/ref/models/instances/#django.db.models.Model.save) method, loop over them and call [save()](https://docs.djangoproject.com/en/5.2/ref/models/instances/#django.db.models.Model.save), like this [example snipped]

Django Simple History relies on `post_save` so has the same problem with regular `update()` as with `bulk_update()`.  However, the same solution should work: [bulk_update_with_history](https://django-simple-history.readthedocs.io/en/latest/common_issues.html#bulk-creating-and-queryset-updating).

See [this management command](https://github.com/UCLALibrary/ftva-lab-data/blob/main/ftva_lab_data/management/commands/set_hard_drive_location.py#L33) for recent use.

PR #76, opened 2025-09-02
SYS-1990: Allow password reset

henry-r18:
Implements [SYS-1990](https://uclalibrary.atlassian.net/browse/SYS-1990)

### Acceptance criteria
- [ ] When users forget their passwords, they are able to use a password-reset mechanism without manual resets by an administrator
- [ ] Use Django’s standard password-reset system
- [ ] Enable the application container to send email

### Description
This PR adds the ability for users to reset their password if forgotten. A link to Django's default `password_reset` view is added to the `login.html` template, and email configuration is set up via `settings.py` and environment variables.

### Testing
No automated tests are added. Dev configuration and manual testing assumes access to a Gmail account.

In order to test the reset workflow manually, first set up an [app password](https://support.google.com/mail/answer/185833?hl=en) on your personal Gmail account. Then add your email address and app password as environment variables to `.docker-compose_local_secrets.env`:

```
DJANGO_EMAIL_FROM_ADDRESS={YOUR_EMAIL_ADDRESS}
DJANGO_EMAIL_PASSWORD={YOUR_APP_PASSWORD}
```

Restart the application Docker container in order to reload the environment variables: `docker compose up -d`

In a logged-out browser window, navigate to the application and select the "Forgot your password?" link. Enter the email associated with your Digital Data account, and you should receive an email with a cryptographic URL to reset the password on your account.


README.md
```diff
@@ -235,3 +235,21 @@ Our deployment system is triggered by changes to the Helm chart.  Typically, thi
 * Breaking changes: update major level (e.g., `v1.0.1` to `v2.0.0`)
 
 In addition to updating version in the Helm chart, update the Release Notes in `release_notes.html`.  Put the latest changes first, following the established format.
+
+
+### Configuring email
+To test email functionality as a developer, add the following to your local secrets `env` file:
```
akohler, line 241: Since `docker-compose.yml` explicitly uses `.docker-compose_local_secrets.env`, the documentation should too.

README.md
```diff
@@ -235,3 +235,21 @@ Our deployment system is triggered by changes to the Helm chart.  Typically, thi
...
+
+```
+DJANGO_EMAIL_FROM_ADDRESS=your_email
+DJANGO_EMAIL_PASSWORD=your_email_password
+```
+
+You can override the email host and port as well, as the default variables in the main Django `env` file assume use of Gmail:
```
akohler, line 248: Refer explicitly to `.docker-compose_django.env` here.

charts/values.yaml
```diff
@@ -43,6 +43,11 @@ django:
     db_user: ""
     db_host: ""
     db_port: ""
+    email_backend: ""
```
akohler, line 46: Since this is explicitly set in `settings.py`, no need for a variable here.  Being explicit in `settings.py` means the environment can't override it.

charts/prod-ftvalabdata-values.yaml
```diff
@@ -49,6 +49,12 @@ django:
     db_user: "ftva-digital-data"
     db_host: "p-d-postgres.library.ucla.edu"
     db_port: 5432
+    email_backend: "django.core.mail.backends.smtp.EmailBackend"
```
akohler, line 52: Since this is explicitly set in `settings.py`, no need for a variable here.  Being explicit in `settings.py` means the environment can't override it.

akohler (changes requested):
@henry-r18 This works, but has a couple of rough edges when I tested it.

I went through the reset process twice, and both times received a 404 after my new password was set.  Django tried sending me to `/accounts/profile/`, which is not a valid URL.  My password did get set correctly - I could go to `/` afterwards with no problem - but the error should be avoided.

After pasting in my one-time link, I wound up here:
http://127.0.0.1:8000/accounts/reset/MjM/set-password/
Then following the process led me to each of these URLs, with the final one causing the 404:
http://127.0.0.1:8000/accounts/reset/done/
http://127.0.0.1:8000/accounts/login/
http://127.0.0.1:8000/accounts/profile/

Otherwise, a couple of Helm-related changes:
* `EMAIL_BACKEND` is explicitly set in `settings.py` - which is fine, as we're doing SMTP everywhere.  But the explicit setting means there's no need for related variables in the Helm chart, since we can't override the hard-coded value.
* The other new values in the Helm chart require additional changes.
  * They must be added to `templates/configmap.yaml`.  See the [LBS example](https://github.com/UCLALibrary/LBS/blob/main/charts/templates/configmap.yaml#L20-L24) - which is inconsistent too :( .
  * Since the chart itself is being changed by adding new values, `Chart.yaml` must have a version bump as well.  This, along with a change I'll make in the `gitops_kubernetes` repo when this is deployed, will tell ArgoCD there's a real new chart it needs to retrieve, not just changes to existing values.

See [Goal: Update environment variables only](https://uclalibrary.atlassian.net/wiki/spaces/DEV/pages/83447375/Deploying+To+Kubernetes+-+Developer+Cookbook#Goal%3A-Update-environment-variables-only) for more info, and please let me know if any questions.

PR #77, opened 2025-09-03
SYS-1993: Add `get_record_by_uuid` view and URL pattern

henry-r18:
Implements [SYS-1993](https://uclalibrary.atlassian.net/browse/SYS-1993)

### Acceptance criteria
- [ ] An endpoint at `records/uuid/<str:uuid>` allows retrieval of a Digital Data record by UUID
- [ ] The data is returned as JSON, just the same as with the existing `get_record` view
- [ ] Change the `get_record` endpoint to `records/<int:record_id>`, removing the trailing slash, which currently causes a redirect on every request

### Description
This PR adds a `get_record_by_uuid` view and associated URL pattern. I opted to adapt the existing `transform_record_to_dict` view utility to accept either a record ID or a UUID, rather than adding a separate helper function.

### Testing
I didn't add any automated tests, as the `@basic_auth_required` decorator is already covered by tests against the `get_record` view, and the other functionality is straightforward. Let me know if there are some special cases I should test.


akohler (changes requested):
@henry-r18 I appreciate this:
>  opted to adapt the existing transform_record_to_dict view utility to accept either a record ID or a UUID, rather than adding a separate helper function.

but as noted in the detailed comment, I think this pushes too much responsibility into that simple method.  It also leads to some related refactoring, simplifying it even further by moving record retrieval out of the transform method altogether.  No new / separate helper methods needed, just a clearer separation of responsibility between callers and this utility.

Thanks --Andy

henry-r18:
@akohler Refactoring `transform_record_to_dict` to accept a `SheetImport` object makes a lot of sense--thanks for pointing that out! I went ahead and updated the three callers in `views.py`, as well as one caller in `tests.py`. I think I got all of them, but let me know if I missed something.

I also rebased onto @ztucker4's latest work. Looks like there are now 66 tests, and they all pass.

akohler (approved):
@henry-r18 Thanks for making the changes.  This looks good to me now.  I think the 2 views `get_record()` and `get_record_by_uuid()` are candidates for refactoring to reduce shared code, but I'm currently not seeing a clean-enough way to do that, so I'm deferring that to future us.

Tests all pass, and a quick check via `curl` and `jq` gives the expected data, using a random UUID in my local system:
```
curl -s -u admin:MY_PASSWORD http://127.0.0.1:8000/records/uuid/635558c7-7932-4e62-ba90-8eaaa5b5334b | jq .

{
  "id": 23456,
  "hard_drive_name": "",
  "carrier_a": "AAC882",
  "carrier_a_location": "S217-01C-12B",
  "carrier_b": "AAC883",
  "carrier_b_location": "",
  "hard_drive_barcode_id": "",
  "file_folder_name": "FE2267T_KTLA_Alan_Lidow_Finals",
  "sub_folder_name": "FE2267T_KTLA_Alan_Lidow_2048x1556_24fps_QT_Final",
  "file_name": "NO FILE NAME",
  "inventory_number": "FE2267T",
  [ many more fields snipped],
  "uuid": "635558c7-7932-4e62-ba90-8eaaa5b5334b",
  "media_type": "",
  "status": []
}
```
Merging now, thanks --Andy

@ztucker4 FYI

PR #78, opened 2025-09-04
SYS-1992: Fix issue with default from email

henry-r18:
Fixes [SYS-1992](https://uclalibrary.atlassian.net/browse/SYS-1992) (hopefully)

### Bug description
The “password reset” emails from the Digital Data application come from `webmaster@localhost` instead of `softwaredev-systems@library.ucla.edu`, which is set as the `email_from_address` in the Helm chart production values file. When running locally, the from-address appears as the Gmail account used for testing.

### Likely fix
The `DEFAULT_FROM_EMAIL` [Django setting](https://docs.djangoproject.com/en/5.2/ref/settings/#default-from-email) defaults to `webmaster@localhost` if not defined. This PR adds the setting to `settings.py`, using the existing `DJANGO_EMAIL_FROM_ADDRESS` environment variable. Since no changes are made to the env files, this shouldn't require any changes to the Helm charts.

I'm not 100% sure this will fix the issue in production, as I wasn't able to reproduce the bug locally. However, from the description in the Django docs, this setting (or lack thereof) appears to be the source of the issue.

I think what's different about the LBS application is that it's using `smtplib` directly, rather than using Django's built-in mail handler (see [comments](https://github.com/UCLALibrary/LBS/blob/ef326b5548b15f3b2f9c68c5e8ebeca99ed7bf40/lbs/settings.py#L178) in that project's `settings.py` file). That would explain why the LBS application is behaving differently, using the expected from-address when it sends emails.

[SYS-1992]: https://uclalibrary.atlassian.net/browse/SYS-1992?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


akohler (approved):
@henry-r18 Ah, I think I understand, though some of the details may be hidden in how Google handles SMTP mail authentication vs. UCLA campus systems.  It seems like without this setting, testing this locally should have the same behavior as in production... but since it doesn't, the difference in remote mail systems must explain it.
I agree, per the docs, this should fix the problem, but we won't know until we try.
Merging now, but won't get deployed until I hear from Anthony about the ArgoCD problems.  I'll add release notes for everything pending once that's fixed.

Thanks --Andy

PR #79, opened 2025-09-04
[DIGINF-28] - Make sure this value defaults to an empty string instead of null

cachemeoutside:
https://uclalibrary.atlassian.net/browse/DIGINF-28

Details in ticket


akohler (approved):
@cachemeoutside Thanks for the PR.  Seems reasonable, though a little odd to have to specify a default in `configmap` as well as the `quote` function - but who am I to argue with Helm YAML?

I added release notes and bumped our internal version number for deployment, as we have a few new features & fixes queued for deployment.

Merging now, thanks --Andy

PR #81, opened 2025-09-17
SYS-2005: Refactor dropdown choices to use new models

henry-r18:
Implements [SYS-2005](https://uclalibrary.atlassian.net/browse/SYS-2005)

### Acceptance criteria
- [X] Refactor `file_type`, `asset_type`, `no_ingest_reason`, and `media_type` fields from `SheetImport` into separate models
- [X] Each of these should be related to `SheetImport` similar to `assigned_user`
- [X] All existing data must be preserved in migrations
- [X] Each of these should be exposed in the Admin Console interface, for maintenance of values
- [X] Each of these should preserve the current custom order of choices when displayed in forms.
- [X] Each of these should be included in the JSON / dict returned by `transform_record_to_dict()`
- [X] The output from transform_record_to_dict()should be unchanged.
- [X] Fixtures are included for existing values.

### Description
This PR refactors the previous `CharField` fields on the `SheetImport` model into new, separate models and `ForeignKey` fields that relate to those models. To do this, three migration files are added:

#### Migration 0018
This migration adds the new models `AssetType`, `FileType`, `MediaType`, and `NoIngestReason` via standard migration operations. Then a custom `RunPython` operation creates objects for each of these models representing the choices previously configured on the `CharField` fields. Since the models are ordered by their primary keys, and the objects are created in the order of the original choices, the order of the models should match the custom order.

#### Migration 0019
This migration adds new `ForeignKey` fields, temporarily alongside the prior `CharField` fields. Then a custom operation uses the values in the `CharField` fields to lookup the relationship with the new models for the `ForeignKey` fields. This effectively migrates the existing data over to the new `ForeignKey` fields.

#### Migration 0020
This migration removes the old `CharField` fields, then renames the temporarily named `ForeignKey` fields to substitute in for the old fields.

### Testing
Four new tests are added, bringing the total tests in the suite to 70. For manual testing, I spot checked a hand-full of records to confirm data before and after applying the migrations remains the same. I also checked the JSON output by the `records/{record_id}` endpoint properly serializes the values of the new fields, after changes to the `transform_record_to_dict` view utility.

I added the new fields to the list of fields ignored in the `_get_combined_field_data` function of the `clean_imported_data` management command, in order to get some existing tests to pass. I figured that's okay, as I don't think the new fields are relevant to the tests at issue.

New values for the dropdowns can now be added via the admin console, by creating new objects on the respective models.

Setting and updating values in the fields seems to work just fine.

### Screen Capture
screencap

[SYS-2005]: https://uclalibrary.atlassian.net/browse/SYS-2005?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


henry-r18:
@akohler Just pushed those changes; thanks for the review!

PR #101, opened 2026-04-10
DLSR-272: Allow bi-directional relationship mgmt

henry-r18:
Implements [DLSR-272](https://uclalibrary.atlassian.net/browse/DLSR-272)

### Acceptance criteria
- [X] Relationships can be added and edited in both “outgoing” and “incoming” directions from item detail page

### Description
This PR improves the functionality added in #99 to allow bi-directional relationship management from an item's detail page. The relationship type drop-down now displays both `type` and `reverse_type` from the `RelationshipType` objects, grouped under the headings "Outgoing" and "Incoming" respectively.

When a user selects one of these options, the direction of the relationship is parsed from prefixes encoded in the option's value. Parsing of the direction prefixes is handled by overrides of the `__init__` and `clean` methods on the `RelationshipForm`. The view uses the information parsed from the form to determine whether the item currently in context (i.e. the one being viewed on the detail page) should be set as the `source` or the `target` for the relationship being added or edited. Deleting relationships remains unchanged.

### Testing
1 new test is added to cover the "incoming" direction, and the other tests are updated to reflect the new prefixes on the `relationship_type` choice values. 88 tests for the application in all.

### Screencap
screencap

[DLSR-272]: https://uclalibrary.atlassian.net/browse/DLSR-272?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/ftva_lab_data/forms.py
+++ b/ftva_lab_data/forms.py
@@ -1,6 +1,20 @@
 from django import forms
+from django.core.exceptions import ValidationError
 from django.core.validators import FileExtensionValidator
 from .models import SheetImport, RelationshipType
+
+
+def _parse_relationship_type_choice(value: str) -> tuple[bool, int] | None:
+    """Parse value of `relationship_type` field to determine direction of relationship.
+
+    :param value: Value of `relationship_type` field.
+    :return: Tuple containing a boolean indicating whether the relationship is outgoing
+        and the ID of the `RelationshipType` object, or `None` if the value is invalid.
+    """
+    prefix, remainder = value.split(":", 1)
+    if prefix not in ("outgoing", "incoming") or not remainder.isdigit():
+        return None
+    return (prefix == "outgoing", int(remainder))
 
 
 class ItemForm(forms.ModelForm):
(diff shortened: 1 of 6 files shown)


ftva_lab_data/forms.py
```diff
@@ -1,8 +1,22 @@
 from django import forms
+from django.core.exceptions import ValidationError
 from django.core.validators import FileExtensionValidator
 from .models import SheetImport, RelationshipType
 
 
+def _parse_relationship_type_choice(value: str) -> tuple[bool, int] | None:
+    """Parse value of `relationship_type` field to determine direction of relationship.
+
+    :param value: Value of `relationship_type` field.
```
akohler: Please add an example of what `value` should look like. I assume from the implementation that it's something like `incoming:12345`, with the digits being the object id, but adding that info to this comment makes it clear.

akohler (changes requested):
@henry-r18 This is excellent. The code and comments are clear, other than in `parse_relationship_type_choice()` which I left a comment on.  Tests all pass, nothing unpleasant in the logs.  The UI is clean, everything works correctly, and I couldn't find a way to break it via the UI.

Once that comment is cleaned up, I'll merge & deploy.
Thanks --Andy

henry-r18:
Thanks @akohler! I fleshed out the doc-string on `_parse_relationship_type_choice`. Re-requesting review now.

akohler (approved):
@henry-r18 Thanks, much clearer. Merging now.

PR #102, opened 2026-06-12
DLSR-380: Add view for retrieving filtered record sets

henry-r18:
Implements [DLSR-380](https://uclalibrary.atlassian.net/browse/DLSR-380)

### Acceptance criteria
- [X] A new view and URL route are added to allow consumers to fetch a filtered set of records, using a provided query string and field name
- [X] Serialized response data is updated to include relationship information

### Description
This PR updates the `get_all_records` view to `get_records` and adds optional query parameters for filtering, in addition to the params that provide pagination. Now, the `/records/` URL responds with all records by default, paginated in groups of 100 as before. However, new parameters `query` and `fields` can be included with the request to get a filtered set of records, using `view_utils.get_search_result_items`. Meanwhile, `view_utils.transform_record_to_dict` is updated to add relationship data to the records returned by `get_records`.

In the immediate term, these changes are intended to allow the `ftva_mams_data.generate_metadata` script to retrieve batches of records directly from the Digital Data app, rather than relying on an external CSV file as input. The relationships data in particular will be useful, as the script labels records as `assets` or `tracks` in the final JSON metadata output using those relationships.

For consistency's sake, `ftva_etl.DigitalDataClient` can provide a wrapper for this refactored view, then `ftva_mams_data.generate_metadata` can consumer that wrapper.

### Testing
A couple new tests are added to cover the filtering feature, and existing tests are updated to align with the change to the view function's name. There are now 90 tests overall, all passing.

[DLSR-380]: https://uclalibrary.atlassian.net/browse/DLSR-380?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/ftva_lab_data/urls.py
+++ b/ftva_lab_data/urls.py
@@ -32,7 +32,6 @@ urlpatterns = [
     path("logs/<int:line_count>", views.show_log, name="show_log"),
     path("release_notes/", views.release_notes, name="release_notes"),
     path("records/", views.get_records, name="get_records"),
-    # path("records/filtered/", views.get_filtered_records, name="get_filtered_records"),
     path("records/<int:record_id>", views.get_record, name="get_record"),
     path(
         "records/uuid/<str:uuid>", views.get_record_by_uuid, name="get_record_by_uuid"
--- a/ftva_lab_data/urls.py
+++ b/ftva_lab_data/urls.py
@@ -31,7 +31,8 @@ urlpatterns = [
     path("logs/", views.show_log, name="show_log"),
     path("logs/<int:line_count>", views.show_log, name="show_log"),
     path("release_notes/", views.release_notes, name="release_notes"),
-    path("records/", views.get_all_records, name="get_all_records"),
+    path("records/", views.get_records, name="get_records"),
+    # path("records/filtered/", views.get_filtered_records, name="get_filtered_records"),
     path("records/<int:record_id>", views.get_record, name="get_record"),
     path(
         "records/uuid/<str:uuid>", views.get_record_by_uuid, name="get_record_by_uuid"
--- a/ftva_lab_data/views.py
+++ b/ftva_lab_data/views.py
@@ -432,6 +432,16 @@ def get_records(request: HttpRequest) -> JsonResponse:
     :param request: The HTTP request object.
     :return: JSON response containing records and total count.
     """
+    # Guard against unexpected keys in the query parameters,
+    # while allowing for expected keys to be missing
+    expected_keys = ["query", "fields", "offset", "limit"]
+    if any(key not in expected_keys for key in request.GET.keys()):
+        message = f"Invalid query parameters. Valid keys are: {expected_keys}"
+        return JsonResponse(
+            {"error": message},
+            status=400,
+        )
+
     # All params are optional. If none are provided, all records are returned.
     query: str = request.GET.get("query", "")
     fields_raw = request.GET.get("fields", "")
--- a/ftva_lab_data/views.py
+++ b/ftva_lab_data/views.py
@@ -435,12 +435,20 @@ def get_records(request: HttpRequest) -> JsonResponse:
     # All params are optional. If none are provided, all records are returned.
     query: str = request.GET.get("query", "")
     fields_raw = request.GET.get("fields", "")
-    offset_raw = request.GET.get("offset")
+    offset_raw = request.GET.get("offset", 0)
     limit_raw = request.GET.get("limit", 100)
 
-    # Make sure offset and limit are integers, if provided
-    offset = int(offset_raw) if offset_raw else None
-    limit = int(limit_raw) if limit_raw else None
+    # If provided, make sure offset and limit are integers,
+    # reverting to defaults if they cannot be coerced to int
+    try:
+        offset, limit = int(offset_raw), int(limit_raw)
+    except ValueError:
+        offset, limit = 0, 100
+
+    # Handle negative values for offset or limit,
+    # reverting to defaults if either is negative
+    if offset < 0 or limit < 0:
+        offset, limit = 0, 100
 
     # Split fields into a list of strings if provided
     fields: list[str] = [f.strip() for f in fields_raw.split(",") if f.strip()]
@@ -450,15 +458,12 @@ def get_records(request: HttpRequest) -> JsonResponse:
     except FieldError as e:
         return JsonResponse({"error": str(e)}, status=400)
     total_records = queryset.count()
-    # Apply offset/limit pagination only if either is provided
-    if offset and limit:
-        records = queryset[offset : offset + limit]
-    elif offset:
-        records = queryset[offset:]
-    elif limit:
-        records = queryset[:limit]
-    else:
-        records = queryset
+    # Apply offset/limit pagination
+    # Because of defaults set above,
+    # we can safely assume offset and limit are non-negative integers,
+    # so this slice will always give us the desired result.
+    # E.g. offset=0, limit=100 -> [0 : 0 + 100] = [0:100]
+    records = queryset[offset : offset + limit]
 
     records_data = [transform_record_to_dict(record) for record in records]
     response_data = {
--- a/ftva_lab_data/views.py
+++ b/ftva_lab_data/views.py
@@ -3,6 +3,7 @@ from django.contrib import messages
 from django.contrib.auth import get_user_model
 from django.contrib.auth.decorators import login_required, permission_required
 from django.core.paginator import Paginator
+from django.core.exceptions import FieldError
 from django.http import HttpRequest, HttpResponse, StreamingHttpResponse, JsonResponse
 from django.shortcuts import get_object_or_404, render, redirect
 from django.template.loader import render_to_string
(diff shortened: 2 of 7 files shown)


ftva_lab_data/urls.py
```diff
@@ -31,7 +31,8 @@
     path("logs/", views.show_log, name="show_log"),
     path("logs/<int:line_count>", views.show_log, name="show_log"),
     path("release_notes/", views.release_notes, name="release_notes"),
-    path("records/", views.get_all_records, name="get_all_records"),
+    path("records/", views.get_records, name="get_records"),
+    # path("records/filtered/", views.get_filtered_records, name="get_filtered_records"),
```
akohler: Is this a leftover?  The view doesn't exist, so remove this comment unless this will be added via the next ticket.

ftva_lab_data/views.py
```diff
@@ -418,21 +419,51 @@ def get_record_by_uuid(request: HttpRequest, uuid: str) -> JsonResponse:
...
+    :return: JSON response containing records and total count.
     """
-    # Offset and limit set via query parameters
-    offset = int(request.GET.get("offset", 0))
-    limit = int(request.GET.get("limit", 100))
-    records = SheetImport.objects.all().order_by("id")[offset : offset + limit]
+    # All params are optional. If none are provided, all records are returned.
+    query: str = request.GET.get("query", "")
+    fields_raw = request.GET.get("fields", "")
+    offset_raw = request.GET.get("offset")
```
akohler: The default value of 0 was lost.

ftva_lab_data/views.py
```diff
@@ -418,21 +419,51 @@ def get_record_by_uuid(request: HttpRequest, uuid: str) -> JsonResponse:
...
+    limit_raw = request.GET.get("limit", 100)
+
+    # Make sure offset and limit are integers, if provided
```
akohler: These conversions will raise `ValueError` if `int()` fails.  I know we didn't handle this in the previous code, and this is an API call so we'd never make a mistake with our parameters :) but let's add a defense now by wrapping these 2 lines in a `try` block, ensuring they have the desired default values if a conversion error occurs.

EDIT: You should also handle negative values here.  I think they don't make sense for API usage, so if the caller does send negative values - which are valid `int`s - they probably should be forced to 0, or raise an exception if we want to be strict.  I'm on the fence, but 0 probably is fine.

ftva_lab_data/views.py
```diff
@@ -418,21 +419,51 @@ def get_record_by_uuid(request: HttpRequest, uuid: str) -> JsonResponse:
...
+
+    # Split fields into a list of strings if provided
+    fields: list[str] = [f.strip() for f in fields_raw.split(",") if f.strip()]
+
+    try:
+        queryset = get_search_result_items(query, fields)
+    except FieldError as e:
+        return JsonResponse({"error": str(e)}, status=400)
+    total_records = queryset.count()
+    # Apply offset/limit pagination only if either is provided
```
akohler: I think this can be simplified, because of how `None` is treated in slicing.
From [the docs](https://docs.python.org/3.13/library/stdtypes.html#sequence-types-list-tuple-range):
> The slice of s from i to j is defined as the sequence of items with index k such that i <= k < j.
    If i is omitted or None, use 0.
    If j is omitted or None, use len(s).

```
>>> l = list("abcde")
>>> l
['a', 'b', 'c', 'd', 'e']
>>> offset = None; limit = None
>>> l[offset:limit]
['a', 'b', 'c', 'd', 'e']
>>> offset = 1
>>> l[offset:limit]
['b', 'c', 'd', 'e']
>>> offset = None; limit = 3
>>> l[offset:limit]
['a', 'b', 'c']
>>> offset = 1; limit = 3
>>> l[offset: offset + limit]
['b', 'c', 'd']
```

So I think this can just be collapsed down to:
```
if offset and limit:
    records = l[offset : offset + limit]
else:
    records = l[offset : limit]
...

Of course, test to confirm this is correct :)

File: ftva_lab_data/views.py [hunk 5]
```diff
@@ -418,21 +419,51 @@ def get_record_by_uuid(request: HttpRequest, uuid: str) -> JsonResponse:
...
+    elif limit:
+        records = queryset[:limit]
+    else:
+        records = queryset
+
     records_data = [transform_record_to_dict(record) for record in records]
     response_data = {
         "records": records_data,
-        "total_records": SheetImport.objects.count(),
+        "total_records": total_records,
```
akohler, line 481: Is this correct?  Should the caller know how many records were found before filtering (which is what `total_records` is here), or how many records are being returned, which would be the length of `records_data`?

ftva_lab_data/views_utils.py
```diff
@@ -508,6 +508,24 @@ def transform_record_to_dict(record: SheetImport) -> dict:
         # or "" to avoid serializing None to "None"
         record_data[key] = str(record_data.get(key) or "")
 
+    # Add Relationships data
```
akohler, line 511: I think this is all good, but please confirm by adding to the test data.  `transform_record_to_dict()` is tested in `MetadataTestCase`, which tests that a very simple item can be handled.  As the comments in those tests say, we don't need to test every field - we already cover FK-related objects and UUIDs, the problem areas so far - but relationships are new (for JSON export) so should be tested.  I think the way you're adding specific data explicitly here is just fine, but a test would prove that :)

akohler (changes requested):
@henry-r18 Overall, this is excellent.  I like the design and think this will meet our current needs as well as providing flexibility for future ones.

I've requested some changes, mostly small but intended to defend against errors around the offset/limit parameters.

In addition to the automated tests, I did some quick tests via `curl`, for a pair of records with a relationship between them.  Including just the 

```
curl -s --user me:mypass http://127.0.0.1:8000/records/12345 | jq .
  "outgoing_relationships": [],
  "incoming_relationships": [
    {
      "source_uuid": "63e1203e-0896-4fee-8e0a-047d6efa8c3d",
      "relationship_type": "hasTrack"
    }
  ]

curl -s --user me:mypass http://127.0.0.1:8000/records/2 | jq .
  "outgoing_relationships": [
    {
      "target_uuid": "a15c5ac9-3884-4f75-bc04-3c27109a4b97",
      "relationship_type": "hasTrack"
    }
  ],
  "incoming_relationships": []
```

I also tested a filter with limits & offset:
```
curl -s --user me:mypass "http://127.0.0.1:8000/records/?carrier_b=AAC256&offset=3&limit=2"
```
and confirmed it contains 2 records, with correct data.

ftva_lab_data/views.py
henry-r18: @akohler I think the original intent here was to give the caller the total count of all records, for pagination purposes. The caller could specify pages of say 1000 using `?limit=1000`, then figure out how many pages it would need to request to exhaust `total_records`. Am I thinking about that right?

akohler: @henry-r18 That was the original intent, when the method returned all records: the caller had to know how to page through them.  However, this revised method returns filtered records, which *might* be all records but probably isn't.

For example, this franken-command (edit the `user` credentials) retrieves 22 records, pretty-prints the JSON, searches that JSON for those two fields, sorts the results and then provides unique counts.  If you don't have `jq`, it's a great tool: https://jqlang.org/
```
curl -s --user me:mypass "http://127.0.0.1:8000/records/?carrier_b=AAC256&offset=1&limit=22" \
| jq . \
| egrep "hard_drive_location|total_records" \
| sort \
| uniq -c

     22       "hard_drive_location": "217",
      1   "total_records": 29714
```

Each of the 22 records has a `hard_drive_location` field, so they're counted, but the `total_records` (which occurs once, outside the `records` list), shows a total of all the records in the database.  That's not helpful for paging.

I've missed something as well.  Returning `total_records` with the number of records returned is redundant, since the caller can just check the length of `records`.

What we need is the total number of records after filtering, when the caller has supplied offset/limit to get just a subset of that total, or when the limited set is smaller than all the records found by the search.  If I search for something with a lot of results, and don't get them all either by choice or limitations within the API, I want to know about it and know how to get the rest.

henry-r18: @akohler Ahh, I see now...sorry for the confusion! I'm experiencing some brain-fry, so will return to this on Monday.

I'll modify my other comment: not quite ready for re-review!

henry-r18: @akohler Brain-fry lifted for a moment: looks like the syntax of the parameters in the `curl` command isn't quite right.

`?carrier_b=AAC256&offset=1&limit=22` should be `?fields=carrier_b&query=AAC256&offset=1&limit=22`

With these params, I see 22 records returned with `total_records: 999`, which I think is what we want.

However, the view should probably show an error if an unknown param is given. I'll see about that on Monday.

akohler: Adding to this, there seems to be a problem with `limit` in general.

Via direct query, in my db, there are 999 records with `carrier_b="AAC256"`:
```
>>> recs = SheetImport.objects.filter(carrier_b="AAC256")
>>> len(recs)
998
```
But the API seems to return as many as I request, even when I request many more.  In this case, I chose 2345 as an arbitrary number, but anything over 999 returns `limit` records
```
# Searching the formatted JSON for an arbitrary field which is a unique string
curl -s --user me:mypass "http://127.0.0.1:8000/records/?carrier_b=AAC256&offset=0&limit=2345" | jq . | grep "hard_drive_location" | wc -l
2345

# Same thing with 5432, or any number bigger than the 999 there really are
curl -s --user me:mypass "http://127.0.0.1:8000/records/?carrier_b=AAC256&offset=0&limit=5432" | jq . | grep "hard_drive_location" | wc -l
5432
```

I thought it might be the underlying `get_search_result_items()`, but that code hasn't changed, testing via UI gives me the expected results (10 pages of 100, close enough to 999, I didn't count them...), and so does this manual check:
```
>>> from ftva_lab_data.views_utils import get_search_result_items
>>> query="AAC256"
>>> fields=["carrier_b"]
>>> items = get_search_result_items(query, fields)
>>> len(items)
999
```
This is with the updated code, the current as of now.

So something in the `limit` handling is causing a problem, and should be reviewed along with finding the best approach to pagination for cases where it is needed.

In fact, filtering on `carrier_b` at least is not working with offset/limit.

```
curl -s --user me_mypass "http://127.0.0.1:8000/records/?carrier_b=AAC256&offset=0&limit=3" | jq '.records[] .id, .records[] .carrier_b'
2
3
4
""
""
""
```
This gets the first 3 records, supposedly matching that condition, but the results are just the first 3 records in the database.  That `jq` filter is showing the values of 2 fields in the `records` array, but same confirmation with `grep`:
```
curl -s --user my:mypass "http://127.0.0.1:8000/records/?carrier_b=AAC256&offset=0&limit=3" | jq . | egrep '"id"|"carrier_b"'
      "id": 2,
      "carrier_b": "",
      "id": 3,
      "carrier_b": "",
      "id": 4,
      "carrier_b": "",
```

So, some research and fixes needed.  Next week!
Thanks --Andy

akohler: @henry-r18 That brain-fry is obviously contagious, and I must have given it to you.  Of course I get wrong results with wrong parameters.... and currently get right results with right parameters!

That said, please add guards (next week) to the endpoint which thrown an error when unexpected parameters are passed.  Since all are optional - we do want to be able to retrieve all records - the behavior I got was [technically correct](https://www.youtube.com/watch?v=hou0lU8WMgo).  But I think the caller should get slapped, as I should have been, when passing unsupported parameters.

Thank you! --Andy

akohler (commented):
@henry-r18 The code is all looking good now.  Tests make sense and all pass.
Please add release notes and bump version so this can get deployed, and I'll merge.
Thanks --Andy

akohler (approved):
@henry-r18 Thanks, merging now.

